PR #4, opened 2015-01-06
Update other_sensor_defs.c

szabolcszijarto:
It works better this way...


dlech:
Yes, definitely better this way. Thank you!

PR #6, opened 2015-02-03
Fix incorrect 'driver_name' attribute values for EV3 UART sensors

in4lio:
Also `EV3_UART_SENSOR_NAME(suffix)` macros has been added into `ev3_uart_sensor.h`.

### DIFF
--- a/sensors/ev3_uart_sensor.h
+++ b/sensors/ev3_uart_sensor.h
@@ -18,6 +18,8 @@
 
 #include <lego_sensor_class.h>
 
+#define EV3_UART_SENSOR_NAME(type_id) "lego-ev3-uart-" type_id
+
 /**
  * struct ev3_uart_sensor_info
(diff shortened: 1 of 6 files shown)


sensors/ev3_uart_sensor.h
```diff
@@ -18,6 +18,8 @@
 
 #include <lego_sensor_class.h>
 
+#define EV3_UART_SENSOR_NAME(suffix) "lego-ev3-uart-" suffix
```
dlech: The suffix is actually the "type ID" of the sensor, so can we replace `suffix` with `type_id`?

dlech:
This looks great. Just the one nit-pick (about suffix/type_id). Also if you could squash the commits and force push, that would be better than having 2 or 3 commits for one change.

in4lio:
Done.

dlech:
Wonderful! Thank you very much!

PR #9, opened 2015-05-04
Updated LED colors for LEGO EV3 Color Sensor

donwojtallo:


sensors/ev3_uart_sensor_defs.c
```diff
@@ -35,7 +35,7 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
 		.mode_info	= {
 			[0] = {
 				/**
-				 * @description: Reflected light<br />Red LED on
+				 * @description: Reflected light<br />LED color: red
```
dlech: Can we make the LED color descriptions consistent?

I think it is better with a colon (`:`) like you have used here, but it does not match the other descriptions. So either add a colon to the rest of leave this as it was.

sensors/ev3_uart_sensor_defs.c
```diff
@@ -46,7 +46,7 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
 			},
 			[1] = {
 				/**
-				 * @description: Ambient light<br />Red LEDs off
+				 * @description: Ambient light<br />Blue LEDs on
```
dlech: It is probably worth noting here that the blue LED is not at full intensity, but is rather dimly lit.

sensors/ev3_uart_sensor_defs.c
```diff
@@ -109,7 +109,7 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
 				 * the EV3 brick. As a result, the sensor will time out and reset.
 				 *
 				 * @name_footnote: [^cal-mode]
-				 * @description: Calibration ???<br />All LEDs rapidly cycling, appears white
+				 * @description: Calibration ???<br />Red LED flashing every 4 seconds, then goes continous
```
dlech: Interesting. I don't think I ever really tried this mode since the sensor resets itself after a few seconds.

donwojtallo: I don't know if it is important, but it resets after a random period of time each time I try. It's around 4-20 blinks, but I'm not sure.

sensors/ev3_uart_sensor_defs.c
donwojtallo: I was trying to make it consistent, just forgot to change this one. I'll fix it.

sensors/ev3_uart_sensor_defs.c
donwojtallo: Are you sure? The brightness seems to be the same as of the red light. And there is no other mode with blue color... Except modes with white, which is combination of dimmed rgb (I'm guessing).

dlech: Yes, I am sure. I did some research...

http://bricks.stackexchange.com/questions/4898/light-on-color-sensor-is-blue

dlech:
Thank you!

PR #12, opened 2015-05-21
Fix polarity comments in servo and taho motors

ddemidov:
The comments were outdated by 36a91aa494a6d126f80837a8f1229d5e4c330fba.


dlech:
Thanks for catching that.

PR #13, opened 2015-06-25
fixed NxtServo and NxtMMX doc

in4lio:


dlech:
Thanks!

in4lio:
Not at all) Will EV3SensorMUX and NSX2020 (NXT Sensor Multiplexer) port names have i2c address field like NxtNMX?

dlech:
The EV3SensorMUX and NSX2020 *should* include the i2c address in the port name. If they don't, it is a bug and you should open an issue for it.

in4lio:
I have no these two)
```
int ms_nxtmmx_probe_cb(struct nxt_i2c_sensor_data *data)
{
	...
	snprintf(smux->port.port_name, LEGO_PORT_NAME_SIZE, "%s:mux%d",
		 data->port_name, ret);
```
Is the address in `data->port_name`?

dlech:
I just tested them and both have the `:i2c<Y>:` in the port name, so yes, `data->port_name` must already include the address.

in4lio:
Thanks!

PR #15, opened 2015-10-31
Fix inline macros and attributes to alow monolithic kernel build

rhempel:
I could not (and probably should not) push to master - minor changes here to allow a monolithic kernel build.

### DIFF
--- a/include/tacho_motor_class.h
+++ b/include/tacho_motor_class.h
@@ -194,9 +194,4 @@ extern void unregister_tacho_motor(struct tacho_motor_device *);
 
 extern struct class tacho_motor_class;
 
-inline struct tacho_motor_device *to_tacho_motor(struct device *dev)
-{
-	return container_of(dev, struct tacho_motor_device, dev);
-}
-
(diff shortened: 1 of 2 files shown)


include/tacho_motor_class.h
```diff
@@ -194,9 +194,4 @@ extern void unregister_tacho_motor(struct tacho_motor_device *);
 
 extern struct class tacho_motor_class;
 
-inline struct tacho_motor_device *to_tacho_motor(struct device *dev)
```
dlech: FWIW, there are lots of `static inline`s in other header files in the kernel - if we ever need to move this back. But, since it is not used anywhere else, moving it is fine.

dlech:
Looks good to me.

include/tacho_motor_class.h
rhempel: OK, well this works so lets leave it. I should have checked for other `static inline` usage, but it never feels right to declare (not define) in a `.h` file...

PR #16, opened 2015-11-03
part number of PF_M_MOTOR typo in motor doc

BertLindeman:
Probably copied and forgot to change (from NXT motor)


dlech:
Thanks! Glad to know someone is double checking me.

PR #20, opened 2015-12-10
Finish initial support for text_value

rhempel:
Added code per changes discussed earlier. I left the initial copy operation using `snprintf()` to handle the case of the echo passing in a non-terminated string (just in case)

Interestingly, you cannot write to a sysfs text attribute a string with embedded linefeeds - the kernel will break it into multiple writes and only the last string will stick.

We may be able to take advantage of that later when we do a framebuffer daemon (maybe)

### DIFF
--- a/sensors/lego_sensor_class.c
+++ b/sensors/lego_sensor_class.c
@@ -125,7 +125,6 @@
  * .
  * `text_value` (read-only)
  * : Returns a space delimited string representing sensor-specific text values.
- *   Currently the string is limited to 512 bytes.
  * .
  * ### Events
  * .
--- a/sensors/lego_sensor_class.c
+++ b/sensors/lego_sensor_class.c
@@ -2,6 +2,7 @@
  * LEGO sensor device class
  *
  * Copyright (C) 2013-2015 David Lechner <david@lechnology.com>
+ * Copyright (C) 2015      Ralph Hempel <rhempel@hempeldesigngroup.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 as
@@ -121,6 +122,12 @@
  *   see how many values there are. Values with N >= num_values will return an
  *   error. The values are fixed point numbers, so check `decimals` to see if
  *   you need to divide to get the actual value.
+ * .
+ * `text_value` (read-only)
+ * : Returns a string representing sensor-specific text value. The string
+ *   contains space delimited values, is null terminated, and is limited
+ *   to PAGE_SIZE-2 bytes to take into account the NULL terminator. We force
+ *   a trailing linefeed on the return value.
  * .
  * ### Events
  * .
@@ -511,6 +518,24 @@ static ssize_t fw_version_show(struct device *dev, struct device_attribute *attr
 	return snprintf(buf, LEGO_SENSOR_FW_VERSION_SIZE + 2, "%s\n", sensor->fw_version);
 }
 
+static ssize_t text_value_show(struct device *dev, struct device_attribute *attr,
+			      char *buf)
+{
+	struct lego_sensor_device *sensor = to_lego_sensor_device(dev);
+	const char *value;
+ 
+	if (!sensor->get_text_value)
+		return -EOPNOTSUPP;
+ 
+	value = sensor->get_text_value(sensor->context);
+
+	if(IS_ERR(value))
+		return PTR_ERR(value);
+
+	return snprintf(buf, PAGE_SIZE, "%s\n", value);
+}
+
+
 static ssize_t bin_data_read(struct file *file, struct kobject *kobj,
 			     struct bin_attribute *attr,
 			     char *buf, loff_t off, size_t count)
@@ -567,6 +592,7 @@ static DEVICE_ATTR_RO(units);
 static DEVICE_ATTR_RO(decimals);
 static DEVICE_ATTR_RO(num_values);
 static DEVICE_ATTR_RO(bin_data_format);
+static DEVICE_ATTR_RO(text_value);
 /*
  * Technically, it is possible to have 32 8-bit values from UART sensors
  * and >200 8-bit values from I2C sensors, but known UART sensors so far
@@ -596,6 +622,7 @@ static struct attribute *lego_sensor_class_attrs[] = {
 	&dev_attr_decimals.attr,
 	&dev_attr_num_values.attr,
 	&dev_attr_bin_data_format.attr,
+	&dev_attr_text_value.attr,
 	&dev_attr_value0.attr,
 	&dev_attr_value1.attr,
 	&dev_attr_value2.attr,
--- a/sensors/lego_sensor_class.c
+++ b/sensors/lego_sensor_class.c
@@ -124,10 +124,8 @@
  *   you need to divide to get the actual value.
  * .
  * `text_value` (read-only)
- * : Returns a string representing sensor-specific text value. The string
- *   contains space delimited values, is null terminated, and is limited
- *   to PAGE_SIZE-2 bytes to take into account the NULL terminator. We force
- *   a trailing linefeed on the return value.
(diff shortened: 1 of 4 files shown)


user/user_lego_sensor.c
```diff
@@ -63,10 +63,20 @@ static ssize_t address_show(struct device *dev, struct device_attribute *attr,
 	return snprintf(buf, LEGO_NAME_SIZE, "%s\n", sensor->sensor.address);
 }
 
+static ssize_t text_value_store(struct device *dev, struct device_attribute *attr,
+			        const char *buf, size_t count)
+{
+ 	struct user_lego_sensor_device *sensor = to_user_lego_sensor_device(dev);
+ 
+	return snprintf( sensor->text_value, USER_LEGO_SENSOR_TEXT_VALUE_SIZE, buf );
```
rhempel: How is this even working? There's no format specifier!? I'll fix and add to the PR. I swear I tested it.

sensors/lego_sensor_class.c
```diff
@@ -122,6 +123,12 @@
  *   error. The values are fixed point numbers, so check `decimals` to see if
  *   you need to divide to get the actual value.
  * .
+ * `text_value` (read-only)
```
dlech: This is the documentation that will appear on the website. 

> is null terminated, and is limited
   to PAGE_SIZE-2 bytes to take into account the NULL terminator. We force
   a trailing linefeed on the return value.

This is assumed for all sysfs attributes and will be confusing for non-kernel hackers. Let's just omit the bit of text I quoted here.

The really important bits here are

* The value is sensor-specific.
* There MAY be more than one "value" and if so they are space-separated.

user/user_lego_sensor.c
dlech: Also, no spaces before/after parentheses.

user/user_lego_sensor.c
```diff
@@ -107,6 +117,12 @@ static const struct attribute_group *user_lego_sensor_class_groups[] = {
 	NULL
 };
 
+const char *user_lego_sensor_get_text_value(void *context) {
+	struct user_lego_sensor_device *sensor = context;
+
+        return sensor->text_value;
```
dlech: Indent using tabs, not spaces.

user/user_lego_sensor.c
```diff
@@ -138,6 +154,9 @@ int user_lego_sensor_register(struct user_lego_sensor_device *sensor,
 	dev_info(&sensor->dev, "Registered '%s' on '%s'.\n", sensor->sensor.name,
 		 sensor->sensor.address);
 
+	sensor->sensor.context = sensor; 
+        sensor->sensor.get_text_value = user_lego_sensor_get_text_value;
```
dlech: tabs, not spaces

user/user_lego_sensor.h
```diff
@@ -16,9 +16,18 @@
...
+ * limiting it to much less to save memory for each user_lego_sensor 
+ * instance.
+ */
+
+#define USER_LEGO_SENSOR_TEXT_VALUE_SIZE (512-2)
+
 struct user_lego_sensor_device {
 	struct lego_sensor_device sensor;
 	struct device dev;
+        char text_value[USER_LEGO_SENSOR_TEXT_VALUE_SIZE];
```
dlech: tabs, not spaces

user/user_lego_sensor.h
```diff
@@ -16,9 +16,18 @@
 #include <lego.h>
 #include <lego_sensor_class.h>
 
+/* The USER_LEGO_SENSOR_TEXT_VALUE_SIZE can be a maximum of PAGE_SIZE-2
+ * to take into account a trailing newline (added later) and NULL. We are
+ * limiting it to much less to save memory for each user_lego_sensor 
+ * instance.
+ */
+
+#define USER_LEGO_SENSOR_TEXT_VALUE_SIZE (512-2)
```
dlech: I think it would be better to leave out the `-2` here. The newline is not ever included in this buffer, so you don't need to save room for it. Also, a null terminator *is* included in the buffer, so basically, right now you are limiting the buffer to 509 printable characters. 

Whatever you decide here should also be mentioned in the website documentation comments.

dlech: Another thought. You could specify the buffer size in `configfs` and dynamically allocate it. This way, each userspace driver can allocate exactly what it needs and no more. If the size is `0`, then `get_text_value` returns `-EOPNOTSUPP`.

rhempel: I'm going to leave it at 512 bytes for now - there's still an issue with memory management after the file is unlinked. Once that's sorted out I'll fine tune the string allocation

rhempel: I think if we do just a plain `echo "foo" > text_value` then there is automatically a `/n` at the end of the string.

dlech: Then we should strip it off when we store it, otherwise, we will get double newlines when we read it.

rhempel: Which is exactly why I mentioned it - but here's the thing - I can't easily tell if there is supposed to be a triling newline because the user wanted it, or if they forgot to use the `-e` option - I think the better thing to do is to NOT add the `\n` when we copy the string to the buffer for the read operation.

dlech: The user doesn't want it ever (this is a space separated list - there should never be newlines). If the last character before `/0` is `/n`, truncate it by writing over `/n` with `/0`.

dlech: We *DO* want to add the newline when we copy the buffer for the read operation. Other drivers will use string constants and I don't want to have to tack a `/n` on the end of every string.

rhempel: OK, so pull off the last trailing newline if one exists, and force one during the copy. Done.

rhempel:
I updated the files as discussed - hopefully it makes more sense - the driver still segfaults when I unlink the sensor so I'm not sure what's going on there. Probably an order issue when freeing up data structures.

sensors/lego_sensor_class.c
```diff
@@ -122,6 +123,10 @@
...
+ * : Returns a space delimited string representing sensor-specific text values.
+ *   Currently the string is limited to 512 bytes.
```
dlech: This attribute is not the one that is limited to 512 bytes.

user/user_lego_sensor.c
```diff
@@ -43,6 +43,11 @@
  * `bin_data` (write-only)
  * : The written data will be stored and can be read using the corresponding
  *   `bin_data` attribute in the `lego-sensor` class device.
+ * .
+ * `text_value` (write-only)
+ * : The written data will be stored and can be read using the corresponding
+ *   `text_value` attribute in the `lego-sensor` class device.
```
dlech: This is the attribute that is currently limited to 512 bytes.

user/user_lego_sensor.c
```diff
@@ -56,17 +61,35 @@
...
+{
+ 	struct user_lego_sensor_device *sensor = to_user_lego_sensor_device(dev);
+
+	if (count > USER_LEGO_SENSOR_TEXT_VALUE_SIZE)
+		return -EINVAL;
+
+	snprintf(sensor->text_value, USER_LEGO_SENSOR_TEXT_VALUE_SIZE, "%s", buf);
+
+	if (sensor->text_value[count-1] == '\n')
+		(sensor->text_value[count-1] = '\0');
```
dlech: why are there `(` `)` around the assignment expression here?

dlech:
Functionally, everything looks good now. I think I know what is causing the error, but I haven't really looked into it yet. Will do that after this is merged.

rhempel:
One more push to fix the latest issues. @dlech, I appreciate the careful review and the attention to detail on getting the descriptions and intent correct. Any hints for me on what might be wrong on the sensor side or should I just wait?

dlech:
I think it is because we are reusing some stack allocated structs without zeroing them out. I need to look at it though.

PR #22, opened 2016-01-01
Doc update only: mode "rcx-led" should be "led

BertLindeman:
The mode can be `auto tacho-motor dc-motor led raw` not `rcx-led`


dlech:
Thanks!

PR #25, opened 2016-03-01
fixed sensor readout of Hitechnic angle sensor

harmooni:
corresponds to https://github.com/ev3dev/ev3dev/issues/555


sensors/lego_sensor_class.c
```diff
@@ -68,6 +68,7 @@
  * .    - `s16`: Signed 16-bit integer (short)
  * .    - `s16_be`: Signed 16-bit integer, big endian
  * .    - `s32`: Signed 32-bit integer (int)
+ * .	- `s32_be`: Signed 32-bit integer, big endian
```
dlech: please indent using spaces between the `.` and `-` here.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -248,6 +248,19 @@ static void mi_xg1300l_remove_cb(struct nxt_i2c_sensor_data *data)
...
 }
 
+/*
+ *  convert raw data of HiTechnic Angle Sensor to degrees
+ */
+static int ht_angle_scale(void *context, struct lego_sensor_mode_info *mode_info,
+			u8 index, long int *value)
+{
+	
+	*value = (2* mode_info->raw_data[index] + mode_info->raw_data[index+1]);
```
dlech: need a space before the `*` here. Could also be written `(mode_info->raw_data[index] << 1)` instead.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -984,11 +997,13 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 				/**
 				 * @description: Angle
 				 * @units_description: degrees
-				 * @value0: Angle (0 to 180)
+				 * @value0: Angle (0 to 359)
 				 */
 				.name = "ANGLE",
-				.raw_max = 180,
-				.si_max = 180,
+				.data_sets = 2,
```
dlech: This only returns one value here correct? `data_sets` should be 1, not 2, otherwise `value1` will return a meaningless value. Instead, set `data_type` to get the i2c to read two bytes. `'LEGO_SENSOR_DATA_U16` would be appropriate, I think.

dlech:
Thanks for your work on this. Please see my comments and also fix the errors from the travis CI. It is OK if some lines are longer than 80, but there are some really long ones like the the changed if statement that need to be wrapped. Don't wrap documentation comment lines though. They are used to generate some of our web pages and can't handle line breaks in some cases.

WasabiFan:
@dlech it looks like commits were added.

@harmooni GitHub doesn't notify maintainers when you push new commits, so you generally want to comment when you have made changes to let others know.

sensors/lego_sensor_class.c
```diff
@@ -67,7 +67,8 @@
  * .    - `u16`: Unsigned 16-bit integer (ushort)
  * .    - `s16`: Signed 16-bit integer (short)
  * .    - `s16_be`: Signed 16-bit integer, big endian
- * .    - `s32`: Signed 32-bit integer (int)
+ * .	- `s32`: Signed 32-bit integer (int)
+ * .	- `s32_be`: Signed 32-bit integer, big endian
```
dlech: It looks like you are still using tabs here. This is actually parsed as markdown for our website, so it is important to have 4 spaces here.

dlech:
Thanks for making the changes. There is on part in the comments where we still need spaces instead of tabs to make it work with our website. If you can fix that and squash your commits, that would be super.

harmooni:
@WasabiFan thanks for notifying. It's my first time working with GitHub. 

@dlech Sorry, I don't know what I was reading.... Thank you for your feedback. I'm not that familiar with Bit manipulation.

PR #26, opened 2016-03-06
8bit are small for compass values

markosko:
Changed from 8bit to 16bits

### DIFF
--- a/sensors/nxt_i2c_sensor_defs.c
+++ b/sensors/nxt_i2c_sensor_defs.c
@@ -1108,7 +1108,7 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 		},
 		.i2c_mode_info	= (const struct nxt_i2c_sensor_mode_info[]) {
 			[0] = {
-				.read_data_reg	= 0x42,
+				.read_data_reg	= 0x44,
 			},
 		},
 	},
--- a/sensors/nxt_i2c_sensor_defs.c
+++ b/sensors/nxt_i2c_sensor_defs.c
@@ -1103,7 +1103,7 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 				.raw_max = 180,
 				.si_max = 180,
 				.units = "deg",
-				.data_type = LEGO_SENSOR_DATA_S8,
+				.data_type = LEGO_SENSOR_DATA_S16,
 			},
 		},
 		.i2c_mode_info	= (const struct nxt_i2c_sensor_mode_info[]) {
--- a/sensors/nxt_i2c_sensor_defs.c
+++ b/sensors/nxt_i2c_sensor_defs.c
@@ -1096,12 +1096,12 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 			[0] = {
 				/**
 				 * @description: Compass Direction
-				 * @value0: Direction (-180 to 180)
+				 * @value0: Direction (0 to 359)
 				 * @units_description: degrees
 				 */
 				.name = "COMPASS",
-				.raw_max = 180,
-				.si_max = 180,
+				.raw_max = 359,
+				.si_max = 359,
 				.units = "deg",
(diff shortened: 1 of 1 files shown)


dlech:
Have to tested this to see if it actually fixes the problem?

Here are the docs from the HiTechnic website:

> ####Sensor Register Layout
> 
> | Address | Type | Contents |
> |---|----|----|
> | 00 – 07H | chars | Sensor version number |
> | 08 – 0FH | chars | Manufacturer |
> | 10 – 17H | chars | Sensor type |
> | 18 – 3DH | bytes | Not used |
> | 3E, 3FH | chars | Reserved |
> | 40H | byte | Not used |
> | 41H | byte | Mode control |
> | 42H | byte | Heading }  two degree heading |
> | 43H | byte | Heading }  one degree adder |
> | 44, 45H | word | Heading (low byte, high byte) |
> | 46 – 7FH | bytes | Not used |
> 
> The Sensor version number field will report a revision number in the format “²Vn.m” where n is the major version number and m is the revision level. Revision numbers will typically reflect the firmware level. The version number will be used to indicate the hardware level.
> 
> The Manufacturer field will contain “HiTechnc”.
> 
> The Sensor type field will contain “Compass”.
> 
> The Mode control field may be set to 0x00 or 0x43 to define measurement mode or hard-iron calibrate mode.
> 
> The Heading fields will return the currently measured heading during measurement mode operation. During hard-iron calibrate mode operation, this field will jam.
> 
> The heading is obtained by reading location 0x42 to obtain the two degree heading and 0x43 to obtain the one degree adder. The heading can then be computed as
> 
> Heading = (two degree heading *2) + one degree adder.
> 
> If the sign is a problem, then the conventional 16 bit value can be obtained from 0x44 (low byte) and 0x45 (high byte).
> 
> Locn 0x41 is the mode control. Defaults to 0, which is measurement mode. Change to  0x43 to set calibration mode. Must be changed back to 0 when calibration is completed. (If cal. fails, locn. 0x41 will be set to 2 after the host has set it to 0 at the end of the failed calibrate attempt.) 

It looks to me like we should be reading the 16 bit value from 0x44/0x45.

dlech:
It also looks like we could add a calibration mode. Normally, I would implement something like this as a command, but the calibration procedure here is long running, so I think it makes more sense as a mode.

markosko:
0x44 and 0x45 if u want to stick with values in documentation, 0x42 and 0x43 should give value in range from 0 to 360

dlech:
But 0x42/0x43 is not a 16-bit integer. It is an 8 bit integer in 0x42 and one bit in 0x42. We actually introduced a function to handle this case in #25, but since we have the option of using a real 16 bit integer directly with this sensor, I think we should use it.

Also, I can't accept this pull request until someone actually compiles it and confirms that it does indeed fix the problem.

markosko:
So can it be using 16bit number with addresses 0x43 and 0x44 ? If yea I can try compile with these changes but not sure what is needed aside from downloaded ev3 kernel repository for compiling ?

dlech:
See <https://github.com/ev3dev/ev3dev-buildscripts> for how to compile the kernel and get it on your EV3.

rhempel:
I have one of those sensors, so if we get stuck I can build/test after we get the tacho drivers sorted out

markosko:
I sill have today too but must to borrow brick from school, I compiled edited kernel only hadnt brick with sensor home

markosko:
Edited address to take 16bit number from sensor  with values from 0 to 360
Tested and worked

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -1103,12 +1103,12 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 				.raw_max = 180,
```
dlech: need to change these from 180 to 359

dlech: also, can't see it in the diff, but there is a documentation comment line above for value0 that needs to be changed to (0 to 359)

dlech:
Thanks!

PR #27, opened 2016-03-07
Ignore bad signoff in CI

dlech:
Fix travis always failing because of extra line tacked onto diff by github


WasabiFan:
:laughing: Your "fix Travis" PR failed on Travis

PR #30, opened 2016-03-29
Get BrickPi working with tacho motor updates

dlech:


brickpi/brickpi_internal.h
```diff
@@ -23,15 +23,21 @@
...
 #include <lego_port_class.h>
 #include <lego_sensor_class.h>
+#include <tacho_motor_class.h>
+#include <tacho_motor_helper.h>
 
 #include "brickpi.h"
 
+#define BRICKPI_DUTY_PCT_TO_RAW(d) ((d) * 255 / 100)
+#define BRICKPI_DUTY_RAW_TO_PCT(d) ((d) * 100 / 255)
+
```
rhempel: Should these comments go in brickpi.h ?

dlech:
Well, I've added a few dozen extra commits, so this is a bit much to expect anyone to review. I'm going to just merge it and see what happens.

It is still passing the ev3dev-lang-python motor tests, so at least I didn't break that part of the motor drivers.

PR #31, opened 2016-04-28
Add support for mindstorms.com PPS58-Nx pressure sensor

kortschak:
@dlech Please take a look at this.

I'll build and test this of the weekend if I can, but I'd like feedback prior to testing. I'll squash into a single commit when it all looks OK.

### DIFF
--- a/sensors/nxt_i2c_sensor_defs.c
+++ b/sensors/nxt_i2c_sensor_defs.c
@@ -262,6 +262,34 @@ static int ht_angle_scale(void *context,
 	return 0;
 }
 
+/*
+ * mindsensors.com PPS5b-Nx pressure sensor related function
+ *
+ * When the pressure units are changed, the reference pressure value is reset
+ * to zero so to ensure we are returning a meaningful relative value, set the
+ * reference to standard atmospheric pressure in the appropriate units.
+ */
+static void ms_pps58_nx_set_mode_post_cb(struct nxt_i2c_sensor_data *sensor, u8 mode)
+{
+	struct i2c_client *client = sensor->client;
+	int value;
+
+	switch (mode) {
+	case 4:
+		value = 15; /* PSI */
+		break;
+	case 5:
+		value = 1013; /* mbar */
+		break;
+	case 6:
+		value = 101; /* kPa */
+		break;
+	default:
+		return;
+	}
+
+	i2c_smbus_write_word_data(client, 0x47, value);
+}
 
 /**
  * nxt_i2c_sensor_defs - Sensor definitions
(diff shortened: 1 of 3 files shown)


sensors/nxt_i2c_sensor_defs.c
```diff
@@ -262,6 +262,19 @@ static int ht_angle_scale(void *context,
...
+	struct i2c_client *client = sensor->client;
+	int ref;
+
+	if(mode < 4 || 6 < mode)
+		return 0;
+
+	ref = i2c_smbus_read_word_data(client, 0x43);
+	if(ref < 0)
+		return ref;
+	return i2c_smbus_write_word_data(client, 0x47, ref);
```
dlech: Setting to the current value here would have the same effect as the ZERO command, would it not? If the sensor is already under pressure, this could be undesirable. I was thinking that we should set this to standard atmospheric pressure depending on the units.

    switch (mode) {
    case 4:
        value = 15; /* PSI */
        break;
    case 5:
        value = 1013; /* mbar */
        break;
    case 6:
        value = 101; /* kPa */
        break;
    default:
        return 0;
    }

    return i2c_smbus_write_word_data(client, 0x47, value)

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -262,6 +262,19 @@ static int ht_angle_scale(void *context,
 	return 0;
 }
 
+static int ms_pps58_nx_set_mode_post_cb(nxt_i2c_sensor_data *sensor, u8 mode)
+{
+	struct i2c_client *client = sensor->client;
+	int ref;
```
dlech: should be `ret`, not `ref`.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -262,6 +262,19 @@ static int ht_angle_scale(void *context,
...
+
+	if(mode < 4 || 6 < mode)
```
dlech: space after if

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -262,6 +262,19 @@ static int ht_angle_scale(void *context,
...
+		return 0;
+
+	ref = i2c_smbus_read_word_data(client, 0x43);
+	if(ref < 0)
```
dlech: space after if

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -262,6 +262,19 @@ static int ht_angle_scale(void *context,
 	return 0;
 }
 
+static int ms_pps58_nx_set_mode_post_cb(nxt_i2c_sensor_data *sensor, u8 mode)
```
dlech: please add a comment above to separate this from other sensor types. And it would be good to explain what we are doing here too.

sensors/nxt_i2c_sensor_defs.c
kortschak: I'm not sure about this. This would change the behaviour of the ZERO command as well. The notion that I have about this is pretty much the same as how a mass scale is used - the scale is zeroed at first use to get a reference weight.

Either way it should be possible to set the reference to some value (either the current absolute pressure or some arbitrary value). I don't see that this is possible as it stands. What should I add to do that?

sensors/nxt_i2c_sensor_defs.c
kortschak: Was intending "reference". Will change.

sensors/nxt_i2c_sensor_defs.c
kortschak: Followed previously existing style. Changed.

sensors/nxt_i2c_sensor_defs.c
kortschak: Done.

kortschak:
How important are the 80 column test failures?

sensors/nxt_i2c_sensor_defs.c
dlech: This is a mode callback, so will not change the behavior of the ZERO command at all. 

If you want a precise relative pressure reading, then yes, you should use the ZERO command to tare the value. But my thinking here is that we use standard atmospheric pressure here to get you in the ballpark so that you don't *have* to zero the sensor if you don't want to.

dlech: and for advanced uses, such as setting an arbitrary value, there is the `direct` sysfs attribute. You don't have to do anything special for that to work.

sensors/nxt_i2c_sensor_defs.c
dlech: `ret` is pretty standard in the Linux kernel for a return value that can be a value or an error.

sensors/nxt_i2c_sensor_defs.c
kortschak: Fair enough. Done.

dlech:
In the case of the comments, they are used for automatically generated documentation, so wrapping lines can actually break things, so best to leave them as-is.

In the case of the code, try to wrap to 80 characters whenever possible, but if it is just a `;` or `,` past 80 lines, then don't worry about it or if the code is just really unreadable if you wrap lines don't worry about it.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -2582,6 +2610,149 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 			},
 		},
 	},
+	[MS_PPS58_NX] = {
+		/**
+		 * @vendor_name: mindsensors.com
+		 * @vendor_part_number: PPS58-Nx
+		 * @vendor_part_name: Digital Pneumatic Pressure Sensor
+		 * @vendor_website: http://mindsensors.com/index.php?module=pagemaster&PAGE_user_op=view_page&PAGE_id=150
```
dlech: I almost forgot. mindsensors has updated their website since I did this. new link is

    http://www.mindsensors.com/ev3-and-nxt/127-digital-pneumatic-pressure-sensor-for-nxt-or-ev3

sensors/nxt_i2c_sensor_defs.c
kortschak: Done.

sensors/nxt_i2c_sensor_defs.c
kortschak: Superseded by other changes.

sensors/nxt_i2c_sensor_defs.c
kortschak: Done.

kortschak:
Thanks for that, those suggestions are incorporated. I'll ping this when I've had a chance to build and test.

dlech:
Great! Thanks for working on this.

kortschak:
Builds and is tested. Works as expected with one curiosity: invoking the ZERO command sometimes emits a RAW pressure value to the terminal.

Please take another look.

dlech:
> invoking the ZERO command sometimes emits a RAW pressure value to the terminal.

That is very strange. It certainly is not coming from this code though. You should open a new issue and post the terminal output so I can see exactly what you are talking about.

kortschak:
I want to explore it a bit more to see that it's not an artifact of how I was testing. I can do that later today.

kortschak:
I have not been able to repeat this behaviour after reboot. If I see it again, I'll file an issue.

So this PR is ready for review.

dlech:
Thanks again!

kortschak:
Thank *you*. I haven't built a kernel this century - it's much easier than it used to be.

PR #32, opened 2016-05-08
Add NXT Temperature Sensor support

kortschak:
@dlech Please take a look. No testing yet, since I'm wanting to make sure it will be safe before I do.

I don't think this is quite right - the scaling looks to me like I should not need to do what I am doing, but the raw_min and raw_max are u32 and I need negative values. Hence the approach I'm taking.

I'm also not sure about how mode state is retained - the device has a variety of modes (word width and one-shot v continuous), but (sanely) imperial is not supported, so unit mode state needs to be retained somewhere else - I am assuming the lego sensor module handles that, but I can't find any indication of that.

Anyway, when you have time please make some suggestions.

### DIFF
--- a/sensors/nxt_i2c_sensor.h
+++ b/sensors/nxt_i2c_sensor.h
@@ -125,6 +125,7 @@ struct nxt_i2c_sensor_info {
 
 enum nxt_i2c_sensor_type {
 	LEGO_NXT_ULTRASONIC_SENSOR,
+	LEGO_NXT_TEMPERATURE_SENSOR,
 	LEGO_POWER_STORAGE_SENSOR,
 	HT_NXT_PIR_SENSOR,
 	HT_NXT_BAROMETRIC_SENSOR,
@@ -158,6 +159,7 @@ enum nxt_i2c_sensor_type {
  * termination, so no more than 18 chars here!
  */
 #define LEGO_NXT_ULTRASONIC_SENSOR_NAME		"lego-nxt-us"
+#define LEGO_NXT_TEMPERATURE_SENSOR_NAME	"lego-nxt-temp"
 #define LEGO_POWER_STORAGE_SENSOR_NAME		"lego-power-storage"
 #define HT_NXT_PIR_SENSOR_NAME			"ht-nxt-pir"
 #define HT_NXT_BAROMETRIC_SENSOR_NAME		"ht-nxt-barometric"
@@ -191,6 +193,7 @@ enum nxt_i2c_sensor_type {
  */
 #define NXT_I2C_SENSOR_ID_TABLE_DATA				\
 	LEGO_DEVICE_ID(LEGO_NXT_ULTRASONIC_SENSOR),		\
+	LEGO_DEVICE_ID(LEGO_NXT_TEMPERATURE_SENSOR),		\
 	LEGO_DEVICE_ID(LEGO_POWER_STORAGE_SENSOR),		\
 	LEGO_DEVICE_ID(HT_NXT_PIR_SENSOR),			\
 	LEGO_DEVICE_ID(HT_NXT_BAROMETRIC_SENSOR),		\
--- a/sensors/nxt_i2c_sensor_core.c
+++ b/sensors/nxt_i2c_sensor_core.c
@@ -356,6 +356,28 @@ static int nxt_i2c_sensor_remove(struct i2c_client *client)
 
 static struct i2c_device_id nxt_i2c_sensor_id_table[];
 
+
+static int nxt_i2c_sensor_detect_lego_temp(struct i2c_client *client)
+{
+	int ret;
+
+	/*
+	 * Try to set the sensor to 12-bit mode.
+	 */
+	ret = i2c_smbus_write_byte_data(client, 0x01, 0x60);
+	if (ret)
+		return ret;
+
+	/*
+	 * Confirm this is a NXT Temperature Sensor.
+	 */
+	ret = i2c_smbus_read_byte_data(client, 0x01);
+	if (ret < 0 || ret != 0x60)
+		return -ENODEV;
+
+	return 0;
+}
+
 static int nxt_i2c_sensor_detect(struct i2c_client *client,
 				 struct i2c_board_info *info)
 {
@@ -375,6 +397,28 @@ static int nxt_i2c_sensor_detect(struct i2c_client *client,
 		info->platform_data = &apdata->sensor_platform_data;
 	}
 #endif
+
+	/*
+	 * NXT Temperature sensor detection is handled here.
+	 *
+	 * The LEGO Temparature sensor does not conform to the LEGO I2C
+	 * specification and so cannot be handled by the normal auto-detection
+	 * code. However, it is a LEGO-brand sensor that is supported by the
+	 * Mindstorms software, so it is treated here as a special case.
+	 *
+	 * We try to detect the tempearature sensor at address 0x4c, but fall
+	 * through to the standard path in the case that there is a device
+	 * there that does not match expectation in case the address is being
+	 * used by another device.
+	 */
+	if (client->addr == 0x4c) {
+		ret = nxt_i2c_sensor_detect_lego_temp(client);
+		if (ret == 0) {
+			snprintf(info->type, I2C_NAME_SIZE, "%s",
+				LEGO_NXT_TEMPERATURE_SENSOR_NAME);
+			return 0;
+		}
+	}
 
 	/*
 	 * Some sensors can fall asleep during boot, so we try reading twice
@@ -426,8 +470,8 @@ struct i2c_driver nxt_i2c_sensor_driver = {
 	.remove		= nxt_i2c_sensor_remove,
 	.class		= I2C_CLASS_LEGOEV3,
 	.detect		= nxt_i2c_sensor_detect,
-	.address_list	= I2C_ADDRS(0x01, 0x02, 0x03, 0x08, 0x0A, 0x0c, 0x11, 0x18,
-				    0x50, 0x51, 0x52, 0x58),
+	.address_list	= I2C_ADDRS(0x01, 0x02, 0x03, 0x08, 0x0a, 0x0c, 0x11, 0x18,
+				    0x4c, 0x50, 0x51, 0x52, 0x58),
 };
 module_i2c_driver(nxt_i2c_sensor_driver);
 EXPORT_SYMBOL_GPL(nxt_i2c_sensor_driver);
--- a/sensors/nxt_i2c_sensor_defs.c
+++ b/sensors/nxt_i2c_sensor_defs.c
@@ -446,6 +446,63 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 			},
 		},
 	},
+	[LEGO_NXT_TEMPERATURE_SENSOR] = {
+		/**
+		 * @vendor_name: LEGO
+		 * @vendor_part_number: 9749
+		 * @vendor_part_name: MINDSTORMS Temperature Sensor
+		 * @vendor_website: https://education.lego.com/en-us/products/mindstorms-temperature-sensor/9749
+		 * @default_address: 0x4C
+		 */
+		.name		= LEGO_NXT_TEMPERATURE_SENSOR_NAME,
+		.vendor_id	= "LEGO",
+		.product_id	= "Temp.",
+		.num_modes	= 2,
+		.mode_info	= (const struct lego_sensor_mode_info[]) {
+			[0] = {
+				/**
+				 * @description: Continuous measurement
+				 * @value0: Temperature (-55.0 to 128.0)
+				 * @units_description: °C
+				 */
+				.name	= "NXT-TEMP-C",
+				.units	= "C",
+				.raw_min = -14080,
+				.raw_max = 32767,
+				.si_min = -550,
+				.si_max = 1280,
+				.decimals = 1,
+				.data_type = LEGO_SENSOR_DATA_S16_BE,
+			},
+			[1] = {
+				/**
+				 * @description: Continuous measurement
+				 * @value0: Temperature (-67.0 to 262.4)
+				 * @units_description: °F
+				 */
+				.name	= "NXT-TEMP-F",
+				.units	= "F",
+				.raw_min = -14080,
+				.raw_max = 32767,
+				.si_min = -670,
+				.si_max = 2624,
+				.decimals = 1,
+				.data_type = LEGO_SENSOR_DATA_S16_BE,
+			}
+		},
+		.i2c_mode_info	= (const struct nxt_i2c_sensor_mode_info[]) {
+			[0] = {
+				.set_mode_reg	= 0x01,
+				.set_mode_data	= 0x60,
+				.read_data_reg	= 0x00,
+			},
+			[1] = {
+				.set_mode_reg	= 0x01,
+				.set_mode_data	= 0x60,
+				.read_data_reg	= 0x00,
+			},
+		},
+	},
 	[LEGO_POWER_STORAGE_SENSOR] = {
 		/**
 		 * @vendor_name: LEGO
--- a/sensors/other_sensor_defs.c
+++ b/sensors/other_sensor_defs.c
@@ -25,40 +25,6 @@
  * parser can be found in the ev3dev-kpkg repository.
  */
 
-const struct other_sensor_info lm75_defs[] = {
-	[LEGO_NXT_TEMPERATURE_SENSOR] = {
-		/**
-		 * [^addresses]: Valid addresses are 0x48..0x4F (configurable via input pins)
-		 * [^usage]: Sample usage:
-		 *
-		 * Register I2C device:
-		 *
-		 * <pre><code>echo tmp275 0x4C > /sys/bus/i2c/devices/i2c-<port+2>/new_device
-		 * </code></pre>
-		 *
-		 * Finding device class node:
-		 *
-		 * <pre><code>for chip in $(find /sys/class/hwmon -name hwmon*)
-		 * do
-		 *     if [[ "$(cat $chip/device/name)" == "tmp275" ]]
-		 *     then
-		 *         # do whatever
-		 *     fi
-		 * done
-		 * </code></pre>
-		 *
-		 * @vendor_name: LEGO
-		 * @vendor_part_number: 9749
-		 * @vendor_part_name: NXT Temperature Sensor
-		 * @vendor_website: http://education.lego.com/en-us/lego-education-product-database/mindstorms/9749-nxt-temperature-sensor/
-		 * @device_class: [hwmon](https://wiki.archlinux.org/index.php/Lm_sensors) [^usage]
-		 * @default_address: 0x4C
-		 * @default_address_footnote: [^addresses]
-		 */
-		.name		= "tmp275",
-	},
-};
-
 const struct other_sensor_info gpio_pcf857xr_defs[] = {
 	[MS_SENSOR_KIT_PFC8574] = {
 		/**


sensors/nxt_i2c_sensor_core.c
```diff
@@ -356,6 +356,33 @@ static int nxt_i2c_sensor_remove(struct i2c_client *client)
...
+static int nxt_i2c_sensor_detect_lego_temp(struct i2c_client *client)
+{
+	int ret;
+	u8 data[1];
+
+	/*
+	 * Try to set the sensor to 12-bit mode.
+	 */
+	data[0] = 0x60;
+	ret = i2c_smbus_write_i2c_block_data(client, 0x01, 1, data);
```
dlech: it would be simpler to use `i2c_smbus_write_byte_data()` here.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -356,6 +356,33 @@ static int nxt_i2c_sensor_remove(struct i2c_client *client)
...
+	if (ret)
+		return ret;
+
+	/*
+	 * Confirm this is a NXT Temperature Sensor.
+	 */
+	data[0] = 0;
+	ret = i2c_smbus_read_i2c_block_data(client, 0x01, 1, data);
```
dlech: likewise, you can use `i2c_smbus_read_byte_data()` here.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -25,6 +25,36 @@
 #include "ms_ev3_smux.h"
 #include "ms_nxtmmx.h"
 
+/*
+ * convert raw data of LEGO NXT Temperature Sensor to Celsius
+ */
+static int lego_nxt_temp_celsius_scale(void *context,
```
dlech: The lego-sensor class has linear scaling built in, so we shouldn't need any special functions like this unless scaling is non-linear.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -446,6 +476,61 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
...
+		/**
+		 * @vendor_name: LEGO
+		 * @vendor_part_number: 9749
+		 * @vendor_part_name: MINDSTORMS Temperature Sensor
+		 * @vendor_website: https://education.lego.com/en-us/products/mindstorms-temperature-sensor/9749
+		 * @default_address: 0x4C
+		 */
+		.name		= LEGO_NXT_TEMPERATURE_SENSOR_NAME,
+		.vendor_id	= "LEGO",
+		.product_id	= "", /* LMS2012 fakes this with "Temp." */
```
dlech: Please actually use "Temp." as the product ID.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -446,6 +476,61 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
...
+				.si_min = -67,
+				.si_max = 262,
+				.decimals = 1,
+				.data_type = LEGO_SENSOR_DATA_S16_BE,
+				.scale	= lego_nxt_temp_fahrenheit_scale,
+			}
+		},
+		.i2c_mode_info	= (const struct nxt_i2c_sensor_mode_info[]) {
+			[0] = {
+				.read_data_reg	= 0x00,
```
dlech: This may seem redundant since we are already setting the mode (write 0x60 to register 0x01) when initializing the sensor, but they should be added here too. These definitions are actually used for other devices as well that don't use the kernel i2c subsystem. So having `set_mode_reg` and `set_mode_data` here will allow the driver to be used in those cases.

dlech:
> I need negative values

Can we just change `raw_min` and `raw_max` to s32? I'll have to think about this a bit to see how it would affect existing drivers.

> I'm also not sure about how mode state is retained ...  I am assuming the lego sensor module handles that, but I can't find any indication of that.

Correct. The `lego-sensor` class keeps track of the selected mode. Mode 0 will be set when the device is plugged in so that we are in a known state.

> the device has a variety of modes (word width and one-shot v continuous)

Are these actually useful? If they are, we can add more modes in addition to the official LEGO ones.



> No testing yet, since I'm wanting to make sure it will be safe before I do.

Looks safe to me. :wink:

sensors/nxt_i2c_sensor_core.c
kortschak: Thanks, I was looking for something like that, but via the source. Can you point me to good i2c API docs?

sensors/nxt_i2c_sensor_defs.c
kortschak: Would you like me to open an issue to fix LEGO_POWER_STORAGE_SENSOR which also leaves this empty?

sensors/nxt_i2c_sensor_core.c
dlech: Don't know about "good", but...

https://www.kernel.org/doc/Documentation/i2c/smbus-protocol

sensors/nxt_i2c_sensor_defs.c
dlech: I think in that case, it actually needs to be empty for proper matching. I don't have one for testing though.

sensors/nxt_i2c_sensor_core.c
dlech: Also https://www.kernel.org/doc/htmldocs/device-drivers/i2c.html

sensors/nxt_i2c_sensor_defs.c
kortschak: Great, I was hoping that was the case. What I was concerned about was the fact that `min_raw` is `u32`. Does this get cast to `u16` on the basis of `data_type`?

Reading the scaling data on the ev3sources typedata page, I'm confused since it states raw min is -14080.0 and raw max is 32767.0. These don't match the data sheet which give 0xc90 (-880) and 0x7ff (2047) respectively. I'm also curious how the 262.4°F Si Max is handled given the integral nature of `si_max` (below I just truncated - should I be working in x10^decimals there?)

sensors/nxt_i2c_sensor_defs.c
kortschak: Ack.

sensors/nxt_i2c_sensor_core.c
kortschak: Ack.

sensors/nxt_i2c_sensor_defs.c
dlech: > Great, I was hoping that was the case. What I was concerned about was the fact that min_raw is u32. Does this get cast to u16 on the basis of data_type?

No, this is something we need to look into more. We might actually need to make raw min/max s64.

> Reading the scaling data on the ev3sources typedata page, I'm confused since it states raw min is -14080.0 and raw max is 32767.0. These don't match the data sheet which give 0xc90 (-880) and 0x7ff (2047) respectively.

We are using 12-bit data mode, so the last 4 bits are always 0. We aren't shifting the bits >> 4, so our raw values will actually be -880 * 16 = -14080 and 2047 * 16 = 32752 (LEGO must have used 32767 out of habit). 

> I'm also curious how the 262.4°F Si Max is handled given the integral nature of si_max (below I just truncated - should I be working in x10^decimals there?)

We can't do floating point in the kernel, so we have to used fixed decimals, so `si_max` will be 2624 and we will also need to set `decimals = 1` to tell userland that it is x10

dlech: I've thought about this a bit more. I would like to change `raw_max` and `raw_min` to s32. There are only a couple sensors that actually use 32-bit raw data and they are signed. All unsigned cases are 16-bit or less, so will fit in s32.

sensors/nxt_i2c_sensor_defs.c
dlech: Thinking about it a bit more... yes, please open an issue to discuss more.

kortschak: Done #626.

I have one, so I can test this when needed.

sensors/nxt_i2c_sensor_defs.c
kortschak: >We are using 12-bit data mode, so the last 4 bits are always 0. We aren't shifting the bits >> 4, so our raw values will actually be -880 * 16 = -14080 and 2047 * 16 = 32752 (LEGO must have used 32767 out of habit).

Thanks, I elided the shift mentally.

kortschak:
>Are these actually useful? If they are, we can add more modes in addition to the official LEGO ones.

Narrower width might be. The AD conversion is more expensive when wider, so e.g. polling time is reduced at 9 bit. I don't know how important this would be (@gquintanac ?). It could easily be added later.

sensors/nxt_i2c_sensor_defs.c
kortschak: Again, this differs from the Power Storage Sensor. Done here.

kortschak:
I'll have a chance to test this on Wednesday. I'll ping then.

gquintanac:
Hi @kortschak and @dlech , 

It would be desirable to have 12 bits resolution but still it worths a try.

I have prepared a small code to show you my observations. In the file timeT.txt you will see that it takes from 60ms or more to obtain consecutive readings. I think it is because of the file access to "/sys/class/hwmon/hwmon0/temp1_input" to get the readings. I would like to get readings every 20ms if possible.

[measuredT.txt](https://github.com/ev3dev/lego-linux-drivers/files/255536/measuredT.txt)
[timeT.txt](https://github.com/ev3dev/lego-linux-drivers/files/255537/timeT.txt)

`// -*-c-*-
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/time.h>
#include <stdlib.h>
#include <math.h>

#define SENSOR_FILE "/sys/class/hwmon/hwmon0/temp1_input"

float readSensor(void) {
    int fd;
    float val = -1;
    char buf[32];
    fd = open(SENSOR_FILE, O_RDONLY);
    if (fd < 0) {
        printf("Failed to open %s\n", SENSOR_FILE);
        return val;
    }
    if (read(fd, &buf, sizeof(buf)) > 0) {
        val = atof(buf) / 1000;
        //printf("Sensor value = %d\n", val);
    } else {
        printf("Failed to read %s\n", SENSOR_FILE);
    }
    close(fd);
    return val;
}

float timeval_diff(struct timeval *difference,
                   struct timeval *end_time,
                   struct timeval *start_time) {
  struct timeval temp_diff;
  float diff_t_s;

  if(difference==NULL) {
    difference=&temp_diff;
  }

  difference->tv_sec  = end_time->tv_sec  - start_time->tv_sec ;
  difference->tv_usec = end_time->tv_usec - start_time->tv_usec;

  /* Using while instead of if below makes the code slightly more robust. */

  while(difference->tv_usec < 0) {
    difference->tv_usec += 1000000;
    difference->tv_sec  -= 1;
  }
  //printf("%d s + %f ms\n",difference->tv_sec, (float)(difference->tv_usec)/1000);
  diff_t_s =  (float)(difference->tv_sec + (difference->tv_usec)/1000);
  return diff_t_s;
}

int main(void)
{
  FILE *fm,*ft;
  struct timeval earlier;
  struct timeval later;
  float interval;

  float yread, y[100], deltay[100];
  int count_data_read, rem_0 = 0;
  short max_num_data = 20; // num of data values to be collected

  fm = fopen("measuredT.txt", "w+");
  ft = fopen("timeT.txt", "w+");

  // start measuring elapsed time
  if(gettimeofday(&earlier,NULL)) {
    perror("first gettimeofday()");
    exit(1);
  }
  // measure and display the temperature
  count_data_read = 1;
  int finished_reading = 0;
  interval = 0;
  while (finished_reading == 0) {
    yread = readSensor(); //read temperature from sensor
    //yread = low_pass(yread); // to smooth data
      if (count_data_read > 1) {
        y[count_data_read - 1] = yread;
        deltay[count_data_read-2] = y[count_data_read-1] - y[count_data_read-2];
        if (deltay[count_data_read-2] == 0){
          rem_0 = 1;
        } else {
          rem_0 = 0;
        }
      }
        // start measuring elapsed time
        if(gettimeofday(&later,NULL)) {
          perror("second gettimeofday()");
          exit(1);
        }

        interval = interval + timeval_diff(NULL,&later,&earlier);

        // restart measuring elapsed time
        if(gettimeofday(&earlier,NULL)) {
          perror("first gettimeofday()");
          exit(1);
        }
        //write data to files
        if (count_data_read < max_num_data) {
          if (rem_0 == 0) {
            system("clear");
            printf("\nTemperature read from Sensor: %f\n", yread);
            printf("Time: %f in s\n", interval/1000);
            fprintf(fm, "%f\n", yread);
            fprintf(ft, "%f\n", interval/1000);
            count_data_read++;
          }
        } else {
          finished_reading = 1;
          printf("Finished collecting data\n");
          sleep(1);
        }
  }
  fclose(fm);
  fclose(ft);
  return 0;
}
`

kortschak:
The changes boot fine, but I don't see anything in `/sys/class/lego-sensor`. `dmesg` reports the same as previously:

```
[  199.076140] i2c-legoev3 i2c-legoev3.6: registered on input port 4
[  199.078892] lego-port port3: Added new device 'in4:nxt-i2c-host'
```

As an aside, assuming I get this working at some point. What should I do with the [I2C/Other entry for the device](https://github.com/ev3dev/lego-linux-drivers/blob/master/sensors/other_sensor_defs.c#L29)?

kortschak:
Sorry, the implied question is how do I go about debugging this?

dlech:
> I would like to get readings every 20ms if possible.

These are your choices:

selection_009

The question I have is: what temperature are you measuring that can change so fast that you could see a 0.5℃ change in 20-30ms?

dlech:
> Sorry, the implied question is how do I go about debugging this?

Add `printk` statements to look at what you need to see.

    printk("I'm debugging %d\n", variable);

Then check `dmesg` for the messages. I usually use a serial console on input port 1 for watching kernel messages. This is the only way to catch crashes, but it requires additional hardware (USB/Serial converter) and some extra configuration. Opening a second terminal window and running `dmesg -w` should work just as well as long as you aren't crashing the kernel.

dlech:
The best place to start would be in the detect function to see what is going on there.

dlech:
> What should I do with the I2C/Other entry for the device?

Delete it.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -356,6 +356,30 @@ static int nxt_i2c_sensor_remove(struct i2c_client *client)
...
+	ret = i2c_smbus_write_byte_data(client, 0x01, 0x60);
+	if (ret)
+		return ret;
+
+	/*
+	 * Confirm this is a NXT Temperature Sensor.
+	 */
+	ret = i2c_smbus_read_byte_data(client, 0x01);
+	if (ret < 0)
+		return ret;
```
dlech: This should return `-ENODEV` instead of the actual error.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -356,6 +356,30 @@ static int nxt_i2c_sensor_remove(struct i2c_client *client)
...
+	if (ret != 0x60)
```
dlech: can be combined with the previous if statement

sensors/nxt_i2c_sensor_defs.c
dlech: Don't forget to add `"Temp." here.

dlech:
> but I don't see anything...

I just noticed that you missed adding

    LEGO_DEVICE_ID(LEGO_NXT_TEMPERATURE_SENSOR),

to 

    #define NXT_I2C_SENSOR_ID_TABLE_DATA

in `sensors/nxt_i2c_sensor.h`. This is most likely why it is not working.

sensors/nxt_i2c_sensor.h
```diff
@@ -158,6 +159,7 @@ enum nxt_i2c_sensor_type {
  * termination, so no more than 18 chars here!
  */
 #define LEGO_NXT_ULTRASONIC_SENSOR_NAME		"lego-nxt-us"
+#define LEGO_NXT_TEMPERATURE_SENSOR_NAME		"lego-nxt-temp"
```
dlech: In <https://github.com/kortschak/lego-linux-drivers/blob/fb80a2581826838a11c9d3773a201298c48112ac/sensors/nxt_i2c_sensor.h>, it looks like there are too many tabs here.

kortschak: Done.

(I already miss a language that uses space to align).

kortschak:
>This is most likely why it is not working.

:) Yes, undoubtedly. Thank you.

sensors/nxt_i2c_sensor_core.c
kortschak: Done.

sensors/nxt_i2c_sensor_core.c
kortschak: Done.

kortschak:
>Add `printk` statements to look at what you need to see.

>>`printk("I'm debugging %d\n", variable);`

Excellent, just what I was looking for.

sensors/nxt_i2c_sensor_defs.c
kortschak: Will do now. I wanted to get things working first.

kortschak:
Tested working.

PTAL

kortschak:
I'll squash when you give the go-ahead.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -377,6 +399,18 @@ static int nxt_i2c_sensor_detect(struct i2c_client *client,
 #endif
 
 	/*
+	 * NXT Temperature sensor detection is handled here.
+	 */
+	if (client->addr == 0x4c) {
+		ret = nxt_i2c_sensor_detect_lego_temp(client);
+		if (ret)
```
dlech: Hmm... If there are ever any other sensors with this address, I thing we would want this to be...

    if (ret == 0) {
        snprintf(info->type, I2C_NAME_SIZE, "%s",
                 LEGO_NXT_TEMPERATURE_SENSOR_NAME);
        return 0;
    }

And then let flow fall though to the regular detection if it is not a temperature sensor.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -377,6 +399,18 @@ static int nxt_i2c_sensor_detect(struct i2c_client *client,
 #endif
 
 	/*
+	 * NXT Temperature sensor detection is handled here.
```
dlech: Please add a sentence that explains the reason we are doing this. Technically, this driver is only for I2C devices that provide vendor and product id as per LEGO's unofficial spec. But we are making an exception since this is a LEGO brand sensor.

sensors/nxt_i2c_sensor_core.c
dlech: Or maybe

    if (ret < 0 && ret != -ENODEV)
        return ret;

dlech:
Also, is there a commit for the change of `raw_min` and `raw_max` to `s32`?

kortschak:
>Also, is there a commit for the change of `raw_min` and `raw_max` to s32?

Sorry, I thought you were going to do that. I'm happy to, but I think it should be separate from this PR since it will potentially be affecting other code. Also, `si_min` and `si_max` should be signed as well if the raw values are changed (the si values can be negative here, so presumably the same argument holds for that field).

If you want me to do it, I'll either do it here as a commit preceding the temperature sensor commits unsquashed or as a separate PR.

dlech:
> Sorry, I thought you were going to do that.

I have plenty of other things to do. :wink: Adding as a separate commit in this pull request is great.

And yes, `si_min` and `si_max` need to be signed as well. I thought we were using negative values some places already, but I guess not.

sensors/nxt_i2c_sensor_core.c
kortschak: I don't see how the second one works since by the time we have entered `nxt_i2c_sensor_detect_lego_temp` we are committed to the belief that the device is the temperature sensor. The only return values we can get is a failure from the write, `-ENODEV`or 0. A failure from the write seems like something that would indicate more than just "this is not the temp sensor, but it might be something else" since it's at address 0x1, which presumably the vast majority of devices would map.

sensors/nxt_i2c_sensor_core.c
kortschak: Done.

kortschak:
https://github.com/ev3dev/ev3dev/issues/636

sensors/nxt_i2c_sensor_core.c
dlech: You can also get an error from the i2c function if we fail to communicate. The first option might be better though.

kortschak:
PTAL

The value range commit precedes the temperature sensor commit, but because of github's approach to ordering commits in a PR, it looks like it is second.

Closes ev3dev/ev3dev#636.

sensors/ev3_uart_sensor_ld.c
```diff
@@ -232,12 +232,12 @@ struct ev3_uart_port_data {
 	u8 requested_mode;
 	u8 type_id;
 	u8 new_mode;
-	u32 raw_min;
-	u32 raw_max;
+	s32 raw_min;
+	s32 raw_max;
 	u32 pct_min;
```
dlech: We might as well change `pct_*` to `s32` as well while we are at it - although we aren't really using it for anything at the moment.

dlech: wait a minute. In this file, we are using u32 here because these are actually 32-bit floating point values coming from the sensors. Since the kernel doesn't have a floating point type, this is actually more like "raw data", so don't change anything in this file.

kortschak: So revert this commit?

dlech: Yes drop this commit. The shared struct is in <https://github.com/ev3dev/lego-linux-drivers/blob/master/include/lego_sensor_class.h#L67>, and already uses signed values.

kortschak: Great, thanks.

Done.

dlech:
you have trailing whitespace in your latest push too that needs to be fixed.

kortschak:
Yeah, I saw that. Already fixed. Thanks.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -377,6 +399,28 @@ static int nxt_i2c_sensor_detect(struct i2c_client *client,
...
+	 *
+	 * The LEGO Temparature sensor does not conform to the LEGO I2C
+	 * specification and so cannot be handled by the normal auto-detection
+	 * code. However, it is a LEGO-brand sensor that is supported by the
+	 * Mindstorms software, so it is treated here as a special case.
+	 *
+	 * We try to detect the tempearature sensor on 0x42, but fall through
```
dlech: Where does the number 0x42 fit into this? Did you mean "at address 0x4c"?

sensors/nxt_i2c_sensor_core.c
```diff
@@ -427,7 +471,7 @@ struct i2c_driver nxt_i2c_sensor_driver = {
 	.class		= I2C_CLASS_LEGOEV3,
 	.detect		= nxt_i2c_sensor_detect,
 	.address_list	= I2C_ADDRS(0x01, 0x02, 0x03, 0x08, 0x0A, 0x0c, 0x11, 0x18,
-				    0x50, 0x51, 0x52, 0x58),
+				    0x4C, 0x50, 0x51, 0x52, 0x58),
```
dlech: I'm not sure which would be more consistent, lowercase or upper case. We seem to have both. Lowercase is more common though, I think.

sensors/nxt_i2c_sensor_defs.c
```diff
@@ -446,6 +446,63 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 			},
 		},
 	},
+	[LEGO_NXT_TEMPERATURE_SENSOR] = {
+		/**
+		 * @vendor_name: LEGO
+		 * @vendor_part_number: 9749
+		 * @vendor_part_name: MINDSTORMS Temperature Sensor
+		 * @vendor_website: https://education.lego.com/en-us/products/mindstorms-temperature-sensor/9749
+		 * @default_address: 0x4C
```
dlech: Uppercase C is good here because this becomes documentation on our website.

sensors/nxt_i2c_sensor_core.c
kortschak: Yes. Fixing.

sensors/nxt_i2c_sensor_core.c
kortschak: Fixing.

kortschak:
PTAL

sensors/nxt_i2c_sensor_core.c
```diff
@@ -377,6 +399,28 @@ static int nxt_i2c_sensor_detect(struct i2c_client *client,
...
 
 	/*
+	 * NXT Temperature sensor detection is handled here.
+	 *
+	 * The LEGO Temparature sensor does not conform to the LEGO I2C
+	 * specification and so cannot be handled by the normal auto-detection
+	 * code. However, it is a LEGO-brand sensor that is supported by the
+	 * Mindstorms software, so it is treated here as a special case.
+	 *
+	 * We try to detect the tempearature sensor at address 0x42, but fall
```
dlech: This still says 0x42. This sensor is at 0x4c.

kortschak: Ugh, sorry. Too early in the morning. *Now* it's fixed.

dlech: We better get this man some more coffee. :laughing:

dlech:
Thanks for sticking with this one!

kortschak:
Learning a lot, thanks for the guidance. The next one should be easy.

kortschak:
Updates ev3dev/ev3dev#622.

PR #33, opened 2016-05-10
Add pressure range comments for documentation

kortschak:
Fixes ev3dev/ev3dev#628.

@dlech Please take a look.


dlech:
For issue cross referencing to work, you actually have to say Fixes ev3dev/ev3dev#628 or it will try to link to `#628` in this repository, which does not exist.

Looks good to me. Although I would say `minus` instead of `offset by` with regard to the reference pressure so that you know which direction it is offset.

kortschak:
Done.

PTAL

dlech:
Awesome, thanks.

PR #34, opened 2016-05-12
Update lego_port_class.c

JorgePe:
Add a note to also use I2C  address with set_device attribute for NXT sensors.

### DIFF
--- a/core/lego_port_class.c
+++ b/core/lego_port_class.c
@@ -74,7 +74,7 @@
  * : (write-only) For modes that support it, writing the name of a driver will
  *   cause a new device to be registered for that driver and attached to this
  *   port. For example, since NXT/Analog sensors cannot be auto-detected, you
- *   must use this attribute to load the correct driver. Returns `-EOPNOTSUPP`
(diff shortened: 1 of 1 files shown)


core/lego_port_class.c
```diff
@@ -74,7 +74,7 @@
  * : (write-only) For modes that support it, writing the name of a driver will
  *   cause a new device to be registered for that driver and attached to this
  *   port. For example, since NXT/Analog sensors cannot be auto-detected, you
- *   must use this attribute to load the correct driver. Returns `-EOPNOTSUPP`
+ *   must use this attribute to load the correct driver (as also the I2C address). Returns `-EOPNOTSUPP`
```
dlech: I think you mean `and` instead of `as`. Also, we could clarify that the address is needed for I2C sensors only. And I would like to keep lines under 80 chars.

dlech:
Thanks. This won't show up on the website right away. The website page is generated from the source code comments and I usually don't update it unless I do a kernel release.

PR #36, opened 2016-05-13
Use unscaled values in value range comment

kortschak:
I think I misinterpreted the meaning behind the range in these comments - thinking it was the post-scaled-range. So this fixes that.

@dlech Please take a look.


dlech:
Good catch.

PR #37, opened 2016-05-13
Fix Power Storage Sensor support

kortschak:
* Add special case for detection.
* Fix endianness of data reads.
* Explicitly state product ID.

Issue ev3dev/ev3dev#626.

Tested working.

@dlech Please take a look.

### DIFF
--- a/sensors/nxt_i2c_sensor_core.c
+++ b/sensors/nxt_i2c_sensor_core.c
@@ -356,6 +356,22 @@ static int nxt_i2c_sensor_remove(struct i2c_client *client)
 
 static struct i2c_device_id nxt_i2c_sensor_id_table[];
 
+static int nxt_i2c_sensor_detect_lego_power(struct i2c_client *client)
+{
+	int ret;
+
+	/*
+	 * Confirm there is a device that satisfies the LMS2012 checks.
+	 */
+	ret = i2c_smbus_read_byte_data(client, 0x08);
+	if (ret != 0)
+		return -ENODEV;
+	ret = i2c_smbus_read_byte_data(client, 0x10);
+	if (ret != 0)
+		return -ENODEV;
+
+	return 0;
+}
 
 static int nxt_i2c_sensor_detect_lego_temp(struct i2c_client *client)
 {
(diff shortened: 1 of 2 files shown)


sensors/nxt_i2c_sensor_defs.c
```diff
@@ -513,7 +513,7 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 		 */
 		.name		= LEGO_POWER_STORAGE_SENSOR_NAME,
 		.vendor_id	= "LEGO",
-		.product_id	= "", /* LMS2012 fakes this with "Store." */
+		.product_id	= "Store",
```
kortschak: "Store" confirmed against [LMS2012](https://github.com/mindboards/ev3sources/blob/master/lms2012/lms2012/Linux_AM1808/sys/settings/typedata.rcf#L87). The period in 'Temp." appears to be a place holder for 5 byte identifier.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -356,6 +356,22 @@ static int nxt_i2c_sensor_remove(struct i2c_client *client)
...
 
+static int nxt_i2c_sensor_detect_lego_power(struct i2c_client *client)
+{
+	int ret;
+
+	/*
+	 * Confirm there is a device that satisfies the LMS2012 checks.
+	 */
+	ret = i2c_smbus_read_byte_data(client, 0x08);
+	if (ret < 0)
```
dlech: I think this should be `if (ret != 0)`. It is possible (although very unlikely) that another device could be at this address. However, we would expect this device to have proper vendor and product id at these addresses and return something > 0.

Same with second instance below.

sensors/nxt_i2c_sensor_core.c
```diff
@@ -399,19 +415,28 @@ static int nxt_i2c_sensor_detect(struct i2c_client *client,
...
-	 * through to the standard path in the case that there is a device
-	 * there that does not match expectation in case the address is being
-	 * used by another device.
+	 * We try to detect the sensors at addresses 0x02 and 0x4c, but fall
+	 * through to the standard path when there is a device there that
+	 * does not match expectation in case the address is being used by
+	 * another device.
 	 */
-	if (client->addr == 0x4c) {
+	switch (client->addr) {
```
dlech: I like what you have done here.

sensors/nxt_i2c_sensor_core.c
kortschak: Done. Thanks.

dlech:
One thing I forgot to mention on the last couple of pull requests:

Please add `Issue ev3dev/ev3dev#626` to the end of the commit message. This will cause GitHub to automatically link the commit to the issue. For ev3dev, I prefer this over `Fixes ev3dev/ev3dev#626` which would automatically close the issue.

kortschak:
PTAL

kortschak:
Comment lost in aether:

> FWIW I prefer having issue notes in commit messages, but don't like how github pollutes the linked issue with each force push. Hence the link in the PR rather than the commit. In the past I've waited until the LGTM and added the link immediately prior to merging, but I'm forgetful, and have found that this fails more often than not. Are you OK with polluted issues? If you are then I can include the comments in the commit. Otherwise I'll try to come up with something more robust that works for me.

dlech:
I don't care for the pollution either, but I'm used to git novices that don't know how to squash commits so it hasn't really been a problem in the past.

If your workflow is not squashing any commits until the LGTM, the I think it would be OK. I actually kind of prefer this when there are multiple rounds of review. You can look at the individual commits to see what has changed since the last review.

dlech:
Looks great. Thanks once again!

kortschak:
>If your workflow is not squashing any commits until the LGTM, the I think it would be OK. I actually kind of prefer this when there are multiple rounds of review. You can look at the individual commits to see what has changed since the last review.

Yes, that's our preferred approach at github.com/gonum.

PR #39, opened 2016-06-01
ev3gyro overflow

bmegli:
This is just docs update, in my tests ev3gyro overflows if spinned too many times (as opposed to
getting stuck). If somebody else also tested it maybe there are diffrent
versions of ev3gyro.


dlech:
Yes, I tested this. There are at least 2 different versions of the LEGO EV3 Gyro sensor. I don't know a way to physically tell the difference, but the newer one also has extra modes.

bmegli:
ping, ok so now the docs updated to getting stuck or overflow depending on version

dlech:
Works for me. Thanks!

PR #41, opened 2016-06-05
Uart line discipline bug

bmegli:
Dereferencing through NULL pointer, probably because of reversed order of two lines of code. 

Here lines reversed so that they are not reversed, huh.


dlech:
Thanks!

bmegli:
Ok!

PR #42, opened 2016-06-12
ev3 uart: use receive_buf2

dlech:
This callback is preferred over receive_buf since it returns the count
of the bytes that were actually received.

Issue: ev3dev/ev3dev#671

### DIFF
--- a/sensors/ev3_uart_sensor_ld.c
+++ b/sensors/ev3_uart_sensor_ld.c
@@ -1034,8 +1034,12 @@ static void ev3_uart_receive_buf(struct tty_struct *tty,
 	if (port->closing)
 		return;
 
-	if (count > CIRC_SPACE(cb->head, cb->tail, EV3_UART_BUFFER_SIZE))
+	if (count > CIRC_SPACE(cb->head, cb->tail, EV3_UART_BUFFER_SIZE)) {
+		printk_ratelimited(KERN_ERR
+				   "%s: buffer overrun\n", dev_name(tty->dev));
+		schedule_work(&port->rx_data_work);
 		return;
+	}
 
 	size = CIRC_SPACE_TO_END(cb->head, cb->tail, EV3_UART_BUFFER_SIZE);
 	if (count > size) {


dlech:
This seems to be working well after numerous reboots. @bmegli, does this look good to you?

dlech:
along with adding `console=tty0` to the kernel command line, I haven't had any lockups (yet)

bmegli:
@dlech I am still reading up on the topic, it will take me some time.

I would at least add some `printk` here:
``` C
/*  
 * Check for buffer overrun. We say that we have handled the data even  
 * though we have not. The work queue handler will most likely get an  
 * error and attempt to reconnect the next time we receive new data.  
 * (Buffer capacity is EV3_UART_BUFFER_SIZE - 1)  
 */  
if (count >= EV3_UART_BUFFER_SIZE) {  
	schedule_work(&port->rx_data_work);  
	return count;  
}  
```

so that we can distinguish between overrun and other potential error that leads to bad_message_size or CRC failure or inccorect message.

But still reading up. I don't think we are in a hurry. You have already fixed the dangerous one (kernel panic).

bmegli:
``` C
if (count >= EV3_UART_BUFFER_SIZE) {  
    schedule_work(&port->rx_data_work);  
    return count;  
}  
```

I would be afraid here that if we are supplied data always in packets greater than our BUFFER_SIZE
we will not process anything and always claim to have processed it. I am not saying it is likely in our case.

dlech:
I have added printing a message when there is a buffer overrun.

bmegli:
Ok, I will sum up what I learnt (not a lot)

- bad_message_size or CRC fail my be caused by https://github.com/ev3dev/ev3dev/issues/671 

It is good that we know that. We can add printk there and know if this is causing incorrect messages or something else

- I have pointed you somewhat blindly to `receive_buf2` thinking that it will fix the problem, I am sorry for that

The best documentation for receive_buf2 I have found is here:
http://thread.gmane.org/gmane.linux.kernel/1508869

Those are patches that actually added it to kernel and explained why. Quite recent - 2013
I don't think anything but `n_tty` is using it currently. 

The important fragment there is:
> Line disciplines not using input flow control can continue to set
tty->receive_room to a fixed value and only provide the receive_buf()
method.

I don't think we can use flow control with uart sensors. They just keep on sending data blindly on two lines (no hardware control) and I don't think there is software flow control there (but you may know better)

- so we can probably keep on using `receive_buf` and fixed `receive_room` and resync if we are overrun for some reason, but with some printk so that we at least know about overrun

- the way you have implemented it with `receive_room=65536` may even be a good idea even though we can only accept 255. If we are overrun you just discard excess data and start over with fresh readings

So I would probably revert this commit, add printk to know if we are overrun and this causes bad messages and think how to best solve it.

I am truly sorry for pointing you in the wrong direction.

dlech:
No worries. Learning is good.

So what you are suggesting now is that we replace 

```C
    if (count > CIRC_SPACE(cb->head, cb->tail, EV3_UART_BUFFER_SIZE))
            return;
```

with 

```C
    if (count > CIRC_SPACE(cb->head, cb->tail, EV3_UART_BUFFER_SIZE)) {
            printk_ratelimited(KERN_ERR "%s: buffer overrun\n", dev_name(tty->dev));
            schedule_work(&port->rx_data_work); 
            return;
```

correct?

bmegli:
Yes, then we will know if this was causing bad messages or there is something else.

dlech:
updated

dlech:
oops. fixed `>=` that sneaked in there.

PR #46, opened 2016-06-19
Fix overflow in ftoi and itof functions

dlech:
Issue ev3dev/ev3dev#680


bmegli:
Looks ok for me, will work.

kortschak:
Checked with FATCATLAB 9DOF on EV3. Fixes problem.

PR #49, opened 2016-06-25
Fix defect in calculation of new_speed_sp

rhempel:
- Add parenthesis to force multiplication before division in calculation
  of the new_speed_sp because division binds higher than multiplication.

### DIFF
--- a/ev3/legoev3_motor.c
+++ b/ev3/legoev3_motor.c
@@ -668,8 +668,8 @@ static void update_position(struct legoev3_motor_data *ev3_tm)
 	rampdown_endpoint = ev3_tm->position
 			  + ((ev3_tm->speed * rampdown_time) / (2*MSEC_PER_SEC));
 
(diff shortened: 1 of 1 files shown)


dlech:
Have you tested this? Last time I checked, division is the same as multiplication in order of operations, so this change has no effect unless it is doing something with casting to a different type.

rhempel:
Yes, I have tested this :-) And it does work - before I touched a single line of code I went through the algorithms, did dimensional analysis, and could not figure out how in the heck this was not working.

Then some judicous `printk`s told the tale ...

rhempel:
The Travis check failed due to a bad sign-off - does the signoff happen when the PR gets accepted?

dlech:
The failure is because of a line longer than 80 chars, nothing to do with sign-off. We don't do signoffs, which is why it say "Ignored".

PR #53, opened 2016-07-17
Fix race condition in setting tm->command

rhempel:
- Addresses ev3dev/ev3dev#661
- Addresses ev3dev/ev3dev#687

### DIFF
--- a/include/tacho_motor_class.h
+++ b/include/tacho_motor_class.h
@@ -91,6 +91,7 @@ struct tacho_motor_params {
 	int time_sp;
 	int ramp_up_sp;
 	int ramp_down_sp;
+	enum tacho_motor_command command;
 	enum tm_stop_action stop_action;
 };
 
@@ -108,7 +109,6 @@ struct tacho_motor_device {
 	const struct tacho_motor_ops const *ops;
 	const struct ev3_motor_info const *info;
 	void *context;
-	enum tacho_motor_command command;
 	struct tacho_motor_params params;
 	struct tacho_motor_params active_params;
 	int ramp_start_speed;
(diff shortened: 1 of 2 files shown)


dlech:
The reason `tm->command = cmd;` is where it is is because it would result in an incorrect value if `tm_send_command` failed. i.e. if `tacho_motor_class_start_motor_ramp` returns an error, now `tm->command` is incorrect.

There were a number of other places where we faced similar race conditions with `tm->active_params`. The way I fixed it was by passing `new_params` to `tacho_motor_class_start_motor_ramp` so that it had the new value.

I'm not saying this is the right or best way to do it, but it seemed like the best solution to me when I was fixing the other race conditions.

dlech:
Here is the relevant commit to what I was talking about with `new_params`: https://github.com/ev3dev/lego-linux-drivers/commit/2644e9f90b3cb96c60805d10850222cbf2d2b340

rhempel:
I understand about command being potentially incorrect, but the problem is that we need at least one full ramp delay or else the value of `tm->command` is from the previous motion. That's why we get a very brief period (the first ramp period) where the motor runs in the wrong direction. Either way, this fixes the defect, so is the code acceptable or is there a better way around the problem, perhaps passing the command to `tacho_motor_class_start_motor_ramp()`?

dlech:
I would be afraid that this change as it is would introduce other subtle bugs if motors are not behaving properly. I would rather see it passed as a parameter like `new_params`.

Or perhaps we could even make `tm->command` a field of `struct tacho_motor_params` instead of having it as a separate variable.

Or you could just set `tm->command` back to it's previous value if there is an error.

rhempel:
OK, I'll most likely update the code by making `tm->command` a field of `struct tacho_motor_params` - stay tuned.

rhempel:
Grr - I'll squash the commits before the actual merge. Code is tested and seems to work pretty well. I have only built this for the EV3, so I'm not sure if there are merge issues with other platforms..

dlech:
I like the feel of having `tm->command` in parameters. I think it is an improvement beyond just fixing the bug.

And you can now squash commits via the GitHub web interface (as long as you want it to be a single commit).

PR #54, opened 2016-09-08
ev3: re-add poll(2) support for motor state

kortschak:
@dlech @rhempel Please take a look.

I need to test this - I'll put together a trivial poll waiter and have a play on the weekend and update here with the result.

### DIFF
--- a/ev3/legoev3_motor.c
+++ b/ev3/legoev3_motor.c
@@ -727,6 +727,8 @@ static enum hrtimer_restart legoev3_motor_timer_callback(struct hrtimer *timer)
 	if (ev3_tm->run_to_pos_active)
 		update_position(ev3_tm);
 
+	schedule_work(&ev3_tm->notify_state_change_work);
+
 	return HRTIMER_RESTART;
 }
 
--- a/ev3/legoev3_motor.c
+++ b/ev3/legoev3_motor.c
@@ -678,26 +678,26 @@ static void update_position(struct legoev3_motor_data *ev3_tm)
 	if (ev3_tm->speed_pid.setpoint > 0) {
 		if (rampdown_endpoint > ev3_tm->position_sp) {
 			ev3_tm->speed_pid.setpoint = new_speed_sp;
-			ev3_tm->ramping = 1;
+			ev3_tm->ramping = true;
 		}
 
 		if (ev3_tm->position >= ev3_tm->position_sp) {
 			schedule_work(&ev3_tm->notify_position_ramp_down_work);
 			legoev3_motor_stop(ev3_tm,
 					   ev3_tm->run_to_pos_stop_action);
-			ev3_tm->ramping = 0;
+			ev3_tm->ramping = false;
 		}
 	} else if (ev3_tm->speed_pid.setpoint < 0) {
 		if (rampdown_endpoint < ev3_tm->position_sp) {
 			ev3_tm->speed_pid.setpoint = new_speed_sp;
-			ev3_tm->ramping = 1;
+			ev3_tm->ramping = true;
 		}
 
 		if (ev3_tm->position <= ev3_tm->position_sp) {
 			schedule_work(&ev3_tm->notify_position_ramp_down_work);
 			legoev3_motor_stop(ev3_tm,
 					   ev3_tm->run_to_pos_stop_action);
-			ev3_tm->ramping = 0;
+			ev3_tm->ramping = false;
 		}
 	}
 }
--- a/include/tacho_motor_class.h
+++ b/include/tacho_motor_class.h
@@ -118,6 +118,7 @@ struct tacho_motor_device {
 	int ramp_delta_time;
 	unsigned long ramp_end_time;
 	unsigned long last_ramp_work_time;
+	enum tacho_motor_state oldstate;
 	bool ramping;
 	enum dc_motor_polarity polarity;
 	/* private */
--- a/motors/tacho_motor_class.c
+++ b/motors/tacho_motor_class.c
@@ -457,12 +457,9 @@ ssize_t position_store(struct device *dev, struct device_attribute *attr,
 	return size;
 }
 
-static ssize_t state_show(struct device *dev, struct device_attribute *attr,
-			  char *buf)
+static int tacho_motor_get_state(struct tacho_motor_device *tm)
 {
-	struct tacho_motor_device *tm = to_tacho_motor(dev);
-	int ret, i;
-	size_t size = 0;
+	int ret;
 
 	ret = tm->ops->get_state(tm->context);
 	if (ret < 0)
@@ -470,6 +467,20 @@ static ssize_t state_show(struct device *dev, struct device_attribute *attr,
 
 	if (tm->ramping)
 		ret |= BIT(TM_STATE_RAMPING);
+
+	return ret;
+}
+
+static ssize_t state_show(struct device *dev, struct device_attribute *attr,
+			  char *buf)
+{
+	struct tacho_motor_device *tm = to_tacho_motor(dev);
+	int ret, i;
+	size_t size = 0;
+
+	ret = tacho_motor_get_state(tm);
+	if (ret < 0)
+		return ret;
 
 	for (i = 0; i < NUM_TM_STATE; i++) {
 		if (ret & BIT(i))
@@ -1200,7 +1211,13 @@ EXPORT_SYMBOL_GPL(tacho_motor_notify_position_ramp_down);
 
 void tacho_motor_notify_state_change(struct tacho_motor_device *tm)
 {
-	sysfs_notify(&tm->dev.kobj, NULL, "state");
+	int newstate;
+
+	newstate = tacho_motor_get_state(tm);
+	if (newstate != tm->oldstate) {
+		sysfs_notify(&tm->dev.kobj, NULL, "state");
+		tm->oldstate = newstate;
+	}
 }
 EXPORT_SYMBOL_GPL(tacho_motor_notify_state_change);
 


dlech:
I think we should check [here](https://github.com/kortschak/lego-linux-drivers/blob/34895f2b9f670f29ef8c678395a4bab956fc1e33/motors/tacho_motor_class.c#L1203) that the state actually changed so that we don't notify when the state has not really changed. For example, if you tell the motor to stop and then tell it to stop again, we only need to be notified the first time.

dlech:
Also, ramping is done in the tacho_motor_class.c driver. The flag is added [here](https://github.com/kortschak/lego-linux-drivers/blob/34895f2b9f670f29ef8c678395a4bab956fc1e33/motors/tacho_motor_class.c#L472). So, we need to call `tacho_motor_notify_state_change()` whenever the ramping state changes as well.

kortschak:
I wasn't sure about that change. It seems though that it would be more sensible to do it [here](https://github.com/kortschak/lego-linux-drivers/blob/34895f2b9f670f29ef8c678395a4bab956fc1e33/ev3/legoev3_motor.c#L805) in `legoev3_motor.c`.

dlech:
Its not clear which change you are referring to. What exactly would you add to `legoev3_motor_get_state`? IIRC, this function is currently only called in response to reading the sysfs attribute.

dlech:
And as general advice, the more you can do in `tacho_motor_class.c`, the better since it is shared with all drivers whereas any changes to `legoev3_motor.c`  will have to be adapted to the other motor drivers. Some hardware specific differences can't be helped, but anything that is not hardware dependent should go in `tacho_motor_class.c`.

kortschak:
Sorry, I was referring to the change in `state_show` you wanted. I made the change in `legoev3_motor_get_state`, but I can back that out to where you suggested. At the moment, `legoev3_motor_get_state` is the only function that gets called as a `get_state` (all the non-ev3 devices defer to ev3's implementation).

kortschak:
BTW The change I made is in the current commit.

ev3/legoev3_motor.c
```diff
@@ -819,6 +830,9 @@ static int legoev3_motor_get_state(void *context)
 			state |= BIT(TM_STATE_OVERLOADED);
 	}
 
+	if (state != original)
+		tacho_motor_notify_state_change(&ev3_tm->tm);
```
dlech: State does not change here. `legoev3_motor_get_state()` is translating `STATE_*` to `TM_STATE_*` flags. These are two different enums and one is used as bit flags, so comparing `state` to `ev3_tm->state` (via `original`) is not valid anyway.

dlech:
> Sorry, I was referring to the change in `state_show` you wanted.

The change does not need to be made in `state_show`, that is just where the result of the change is seen. You need to find where `tm->ramping` is actually set.

dlech:
> all the non-ev3 devices defer to ev3's implementation

Are you sure? See...
* https://github.com/kortschak/lego-linux-drivers/blob/34895f2b9f670f29ef8c678395a4bab956fc1e33/sensors/ms_nxtmmx.c#L538
* https://github.com/kortschak/lego-linux-drivers/blob/34895f2b9f670f29ef8c678395a4bab956fc1e33/brickpi/brickpi_ports_out.c#L353

kortschak:
I'm not sure any longer :)

Thank you, I'll do some more careful spelunking.

kortschak:
I think this is better.

There are two places, where I'm not sure what to do [here](https://github.com/kortschak/lego-linux-drivers/blob/34895f2b9f670f29ef8c678395a4bab956fc1e33/ev3/legoev3_motor.c#L690) and [here](https://github.com/kortschak/lego-linux-drivers/blob/34895f2b9f670f29ef8c678395a4bab956fc1e33/ev3/legoev3_motor.c#L702) - the `sysfs_notify` call is made prior to the `ramping` change because of the call to `legoev3_motor_stop` (if it was actually a stop). Should I call again if there was a change in `ramping`? Also, what is the style guide for `0`|`false`/`1`|`true` use? The attribute set here differs from the rest of the cases I've seen.

What do you want me to do with brickpi? I don't see any obvious way to handle the situation there. I'm happy to leave it.

motors/tacho_motor_class.c
```diff
@@ -342,6 +343,7 @@ static int tm_do_one_ramp_step(struct tacho_motor_device *tm,
 		if (params->command == TM_COMMAND_STOP)
 			return tm->ops->stop(tm->context,
 					     params->stop_action);
+		tacho_motor_notify_state_change(tm);
```
kortschak: This is placed here because `tm->ops->stop(tm->context, params->stop_action))` calls `sysfs_notify`.

dlech:
> Should I call again if there was a change in ramping?

Yes.

> Also, what is the style guide for 0|false/1|true use?

If the data type is `bool`, then use `true`/`false`, otherwise use `0`/`1`. Or change the data type to `bool` if it really should be boolean.



> What do you want me to do with brickpi?

I think the more general solution that would work for everything is to refactor `state_show` in `tacho_motor_class.c` to a new function and have `state_show` and `tacho_motor_notify_state_change` both call this new common function. This way, you can call `tacho_motor_notify_state_change` anywhere the state *might* change instead of trying to figure out a way to test for every possible state change in every motor driver.

kortschak:
>I think the more general solution that would work for everything is to refactor `state_show` in `tacho_motor_class.c` to a new function and have `state_show` and `tacho_motor_notify_state_change` both call this new common function. This way, you can call `tacho_motor_notify_state_change` anywhere the state *might* change instead of trying to figure out a way to test for every possible state change in every motor driver.

sorry, I don't understand this for a few reasons.

* I can't see anywhere that state_show is actually called in the lego drivers; presumably it is called by other kernel code somewhere - I think I need to understand this before I can think about this properly, but a kernel code search does not really help me. I need to know what the HB relationships are between state changes, `get_state` calls, sysfs notifications, poll waits and client reads.

* I think the core though is that I don't understand how a state change caused in, e.g., `legoev3_motor_run_unregulated` is propagated to `tacho_motor_notify_state_change` or `state_show` without an explicit call - which we have now in the first commit here.

* In brickpi, there is no specific code to change the `state` or `ramping` members - this is what I was referring to in https://github.com/ev3dev/lego-linux-drivers/pull/54#issuecomment-245782812 where I talk about deferring to ev3 - the function members are set in brickpi, but the actual code to drive the motors is in ev3, where the changes I've made above should catch the motor state changes.

dlech:
> can't see anywhere that state_show is actually called in the lego drivers

It is called when you read the `state` sysfs attribute. (Which also happens to be the file we are going to be polling)

The reason you don't find it by searching is that it is used in a macro: https://github.com/ev3dev/lego-linux-drivers/blob/2008a045d0a5fccccf6218c424c71b372fcb2fb1/motors/tacho_motor_class.c#L1031

dlech:
> think the core though is that I don't understand how a state change caused in, e.g., legoev3_motor_run_unregulated is propagated to tacho_motor_notify_state_change or state_show without an explicit call - which we have now in the first commit here.

We will still have the call to `tacho_motor_notify_state_change` as you have it now, but instead of adding `state_changed = ev3_tm->state != STATE_RUNNING;` and calling `tacho_motor_notify_state_change` conditionally based on `state_changed`, we call `tacho_motor_notify_state_change` unconditionally.

Then in the `tacho_motor_notify_state_change` we call the new `tacho_motor_get_state` function.

```
int tacho_motor_get_state(struct tacho_motor_device *tm)
{
	int ret;

	ret = tm->ops->get_state(tm->context);
	if (ret < 0)
		return ret;

	if (tm->ramping)
		ret |= BIT(TM_STATE_RAMPING);

	return ret;
}
```

This function shares code with `state_show`, so `state_show` can be modified to call this function instead of duplicating code.

`tacho_motor_notify_state_change` will look something like...

```
void tacho_motor_notify_state_change(struct tacho_motor_device *tm)
{
        int newstate;

        newstate = tacho_motor_get_state(tm);
        if (newstate != tm->oldstate) {
                sysfs_notify(&tm->dev.kobj, NULL, "state");
                tm->oldstate = newstate'
        }
}
```

dlech:
> In brickpi, there is no specific code to change the state or ramping members

In `brickpi_out_port_get_state`, we can see that the state depends on:

* `data->motor_enabled`
* `data->hold_pid_ena`
* `tm_pid_is_overloaded(&data->speed_pid)`
* `tm_speed_get(&data->speed) == 0`

So we need to call `tacho_motor_notify_state_change` somehow from any code that affects these conditions.

dlech:
Even if it is not currently the case, the EV3 motor driver *should* be looking at:

* tm_pid_is_overloaded(&data->hold_pid)
* tm_pid_is_overloaded(&data->speed_pid)

So these cases will be shared with both the EV3 and BrickPi motor drivers.

kortschak:
Thanks for that explanation. It's clear now.

kortschak:
Though, I am wondering how I thread in the state that is to be changed to/from. Are you suggesting a new `oldstate` member be added?

dlech:
> Are you suggesting a new oldstate member be added?

Yes, to `struct tacho_motor_device`.

kortschak:
The ev3 cases are now covered, including those in https://github.com/ev3dev/lego-linux-drivers/pull/54#issuecomment-246084851 (due to `legoev3_motor_get_state` via `tacho_motor_get_state` and `get_state`).

I'm not sure how to deal with brickpi though; the brickpi devices don't seem to hold onto a `*tacho_motor_device` to allow the communication needed, instead motor state is held in a heap of members that mirror that type in `brickpi_out_port_data`.

kortschak:
Oh, also there are work queue remnants in ev3 as well, [here](https://github.com/ev3dev/lego-linux-drivers/blob/7c0ea502e9b65e0fd0477afa636c68ca7738aab4/ev3/legoev3_motor.c#L733) and below. Should they remain and should I be using them? (If I should be using them, I need to figure out how to set up a `work_struct`.

ev3/legoev3_motor.c
```diff
@@ -955,6 +963,7 @@ static int legoev3_motor_probe(struct lego_device *ldev)
 	hrtimer_init(&ev3_tm->timer, CLOCK_MONOTONIC, HRTIMER_MODE_REL);
 	ev3_tm->timer.function = legoev3_motor_timer_callback;
 
+	ev3_tm->tm.oldstate = 0;
```
dlech: This is not necessary. The `struct tacho_motor_device` is zeroed when it is allocated, so this will already be 0.

include/tacho_motor_class.h
```diff
@@ -195,6 +196,7 @@ struct tacho_motor_ops {
 	int (*set_hold_Kd)(void *context, int k);
 };
 
+extern int tacho_motor_get_state(struct tacho_motor_device *tm);
```
dlech: I don't see this used anywhere outside of `tacho_motor_class.c`. Why do we need it in the header file?

motors/tacho_motor_class.c
```diff
@@ -1198,9 +1196,29 @@ void tacho_motor_notify_position_ramp_down(struct tacho_motor_device *tm)
 }
 EXPORT_SYMBOL_GPL(tacho_motor_notify_position_ramp_down);
 
+int tacho_motor_get_state(struct tacho_motor_device *tm)
```
dlech: This should be `static` if it is not used anywhere else.

dlech:
> Oh, also there are work queue remnants in ev3 as well, here and below. Should they remain and should I be using them?

It is still called from [here](https://github.com/ev3dev/lego-linux-drivers/blob/7c0ea502e9b65e0fd0477afa636c68ca7738aab4/ev3/legoev3_motor.c#L685).

I don't think you need to be using it from anywhere else.

motors/tacho_motor_class.c
```diff
@@ -464,12 +467,7 @@ static ssize_t state_show(struct device *dev, struct device_attribute *attr,
 	int ret, i;
 	size_t size = 0;
 
-	ret = tm->ops->get_state(tm->context);
-	if (ret < 0)
-		return ret;
-
-	if (tm->ramping)
-		ret |= BIT(TM_STATE_RAMPING);
+	ret = tacho_motor_get_state(tm);
```
dlech: Still need to check for `ret < 0` here and return, otherwise we will get garbage out.

motors/tacho_motor_class.c
dlech: Also need to move this function somewhere before `state_show()` so you don't have to declare the prototype.

dlech:
Also we can't call `tacho_motor_notify_state_change()` directly from an atomic context. That includes inside of an interrupt handler. You can recognize these by the `irqreturn_t` return type. hrtimers are soft interrupts, so they fall into this category as well. They have an `enum hrtimer_restart` return type.

So, for example: `legoev3_motor_timer_callback()` is a hrtimer callback and it calls `update_position()` which in turn calls `legoev3_motor_stop()`. Therefore, we can't call `tacho_motor_notify_state_change()` directly from `legoev3_motor_stop()` as it is.

One way to fix this is to call `schedule_work(&ev3-tm->notify_state_change_work)` instead of calling `tacho_motor_notify_state_change()` to defer the call to when we are no longer in an atomic context.

ev3/legoev3_motor.c
kortschak: Thanks, I wasn't sure about that.

motors/tacho_motor_class.c
kortschak: Done

include/tacho_motor_class.h
kortschak: Removed (the movement of the definition resolved this need)

motors/tacho_motor_class.c
kortschak: Fixed

kortschak:
I'll look at the `schedule_work` change when I'm not trying to organise a 5 year old.

kortschak:
OK, I think that covers the cases where we're in an irq.

dlech:
So have you tested it yet? I think it would be best to make sure it actually does what you were hoping it would do before spending more time on it.

kortschak:
No, no tests yet. Eking out time to work on things uninterrupted is a slow business. I'll be able to find time to do that this weekend. I already have the test rig written, it's just a matter of putting things together.

kortschak:
I've had a quick check and so far it does not appear to do what I expect. The testing program is Go, but should be easily readable as a C program:

```
package main

import (
	"flag"
	"fmt"
	"io"
	"log"
	"os"
	"time"

	"golang.org/x/sys/unix"
)

func main() {
	path := flag.String("path", "", "path to target file")
	timeout := flag.Duration("timeout", time.Second, "poll timeout")
	flag.Parse()
	if *path == "" {
		flag.Usage()
		os.Exit(1)
	}

	f, err := os.Open(*path)
	if err != nil {
		log.Fatalf("failed to open %q: %v (%+[2]v)", *path, err)
	}
	defer f.Close()

	start := time.Now()
	fds := []unix.PollFd{{Fd: int32(f.Fd()), Events: unix.POLLIN}}
	n, err := unix.Poll(fds, int(*timeout/time.Millisecond))

	fmt.Printf("n=%d err=%v delay=%v flags=%016b (POLLIN=%t)\n",
		n, err, time.Since(start), fds[0].Revents, fds[0].Revents&unix.POLLIN != 0)
	io.Copy(os.Stdout, f)
}
```

What I'm looking for then is for a default 1 second delay when I set the motor going forever and invoke the program above on the motor's state attribute - the `poll(2)` call should timeout and POLLIN should be unset.

Instead I see this:

```
robot@ev3dev:~$ echo 50 >/sys/class/tacho-motor/motor1/speed_sp 
robot@ev3dev:~$ echo run-forever >/sys/class/tacho-motor/motor1/command
robot@ev3dev:~$ ./poll -path /sys/class/tacho-motor/motor1/state
n=1 err=<nil> delay=99.358µs flags=0000000000001001 (POLLIN=true)
running
robot@ev3dev:~$ ./poll -path /sys/class/tacho-motor/motor1/state
n=1 err=<nil> delay=1.069439ms flags=0000000000001001 (POLLIN=true)
running
robot@ev3dev:~$ ./poll -path /sys/class/tacho-motor/motor1/state
n=1 err=<nil> delay=93.274µs flags=0000000000001001 (POLLIN=true)
running
```

Any suggestions?

dlech:
[Here](https://github.com/ev3dev/ev3dev-tools/blob/ev3dev-jessie/source/notify/sysfs_notify_monitor.c) is @rhempel's old test program. Perhaps it will help. I think the trick is that `sysfs_notify` only works with `POLLPRI|POLLERR`, not `POLLIN`.

kortschak:
Thanks, I've also found that I need to read the attribute - @rhempel does that there too. I'll play more.

kortschak:
Yay!

```
$ echo run-forever >/sys/class/tacho-motor/motor0/command
$ (sleep 5; echo stop >/sys/class/tacho-motor/motor0/command)& ./poll -path /sys/class/tacho-motor/motor0/state 
[1] 619
n=1 err=<nil> delay=4.954335509s flags=0000000000001010 (POLLPRI=true)

$ (sleep 2; echo run-forever >/sys/class/tacho-motor/motor0/command)& ./poll -path /sys/class/tacho-motor/motor0/state 
[2] 626
n=1 err=<nil> delay=1.957047198s flags=0000000000001010 (POLLPRI=true)
running stalled
[1]   Done                    ( sleep 5; echo stop > /sys/class/tacho-motor/motor0/command )
```

New state for the test program:

```
package main

import (
        "flag"
        "fmt"
        "io"
        "log"
        "os"
        "time"

        "golang.org/x/sys/unix"
)

func main() {
        path := flag.String("path", "", "path to target file")
        timeout := flag.Duration("timeout", 10*time.Second, "poll timeout")
        flag.Parse()
        if *path == "" {
                flag.Usage()
                os.Exit(1)
        }

        f, err := os.Open(*path)
        if err != nil {
                log.Fatalf("failed to open %q: %v (%+[2]v)", *path, err)
        }
        defer f.Close()

        // Dummy read to mark old state.
        f.ReadAt([]byte{0}, 0)

        start := time.Now()
        fds := []unix.PollFd{{Fd: int32(f.Fd()), Events: unix.POLLPRI}}
        n, err := unix.Poll(fds, int(*timeout/time.Millisecond))

        fmt.Printf("n=%d err=%v delay=%v flags=%016b (POLLPRI=%t)\n",
                n, err, time.Since(start), fds[0].Revents, fds[0].Revents&unix.POLLPRI != 0)
        io.Copy(os.Stdout, f)
}
```

It does not currently detect the stalled state. But I suspect that is not possible - I'll have a quick look now.

kortschak:
Yeah, it looks like stalled is not readily fitted in this approach without addition a watchdog-like thing to monitor the motors' states. At the moment calls to `sysfs_notify` are only made when there is programmatic change to the state - which a stall is not.

kortschak:
Is there more to do here? My state-waiting client code works around the inability to `sysfs_notify` for things that are not changed programmatically by busy waiting when `stalled` or `overloaded` states are part of the query, so I'm satisfied with what exists here.

dlech:
I don't really like notifying for some events, but not all. (I have a feeling this will lead to many tech support questions because it does not work as expected.)

But, this is a good start and I know fixing `stalled` and `overloaded` will not be easy. I'll have to think about it a bit.

kortschak:
Agreed on all points. I'll squash this and we can either add more if you have suggestions for me to follow, or you can merge and we leave the ticket open and I'll add "Updates ev3dev\#674.". to the commit message.

kortschak:
If you merge, please wait for me to add appropriate "Fixes" or "Updates" commit comment.

dlech:
Now that this has had some time to percolate, I have a new idea. All of the motor drivers have a polling timer of some sort for things like calculating speed, etc. What if instead of trying to call `tacho_motor_notify_state_change()` from each place that it could possibly change, we call it from the timer? This would take care of the issue of `overloaded` and `stalled` not working and would be less prone to bugs of missing other state changes.

kortschak:
That seems pretty reasonable. I'll be able to take a look at that over the weekend.

ev3/legoev3_motor.c
```diff
@@ -727,6 +729,8 @@ static enum hrtimer_restart legoev3_motor_timer_callback(struct hrtimer *timer)
 	if (ev3_tm->run_to_pos_active)
 		update_position(ev3_tm);
 
+	schedule_work(&ev3_tm->notify_state_change_work);
```
kortschak: Is this essentially all that is needed?

Do you want me to remove the notification calls at the other explicit sites?

dlech: I think so.

You might want to save your previous work in another branch in case this doesn't work out, then delete all of the other calls.

kortschak:
I can't do anything further with this until the follwoing problem is fixed - my previous laptop has died and I am trying to install the build requirements for `build-kernel`. So far `gcc-linaro-arm-linux-gnueabihf-5.2_2015.11` has chewed through ~450MB in failed attempts to download the tar.xz (three failures): the download stalls from aws and then gets kicked off. So I can't build the kernel.

dlech:
Try the old `code-sourcery-toolchain-arm-2011.03` package instead. You will need to create a `local-env` file in the same directory as the `build-kernel` scripts with something like this...

```
if [ "$EV3DEV_KERNEL_FLAVOR" = "ev3" ]; then
    export EV3DEV_TOOLCHAIN=${EV3DEV_TOOLCHAIN-"/opt/arm-2011.03/bin"}
    export EV3DEV_ABI=${EV3DEV_ABI-"arm-none-eabi-"}
fi
```

kortschak:
I've modified the deb so it will install from a local blob and will get the blob from a more reliable network connection later today (I suspect that the stalling is due to the wireless on the new machine being not properly supported by Linux yet - there was a fair bit of jiggery pokery to get it to work at all).

kortschak:
I think we have a winner.

```
$ echo 50 >/sys/class/tacho-motor/motor0/speed_sp
$ echo run-forever >/sys/class/tacho-motor/motor0/command
$ ./poll -path /sys/class/tacho-motor/motor0/state
# manually block motor
n=1 err=<nil> delay=5.755591004s flags=0000000000001010 (POLLPRI=true)
running stalled
```

If you're happy with the current state, I'll squash and then you can take it from there.

dlech:
Can I get you to add a call to `tacho_motor_notify_state_change()` [here](https://github.com/ev3dev/lego-linux-drivers/blob/master/brickpi/brickpi_ld.c#L483).

It looks like adding this for PiStorms/NxtMMX is non-trivial since there isn't a timer that polls motor state already.

kortschak:
I don't see how I can do that; there is no reference to a `tacho_motor_device` in `brickpi_data`.

dlech:
Actually, the call should go [here](https://github.com/ev3dev/lego-linux-drivers/blob/master/brickpi/brickpi_ld.c#L276). But of course, there is no reference in `struct brickpi_out_port_data` either.

Perhaps we should add a timer to the tacho motor class that polls drivers every 100ms for changes in state instead of trying to do something in the individual motor drivers?

dlech:
Another thing we could do instead is make use of the [lego_port_notify_raw_data_func_t notify_raw_data_func](https://github.com/ev3dev/lego-linux-drivers/blob/master/include/lego_port_class.h#L115) callback we use for sensors. Then the tacho motor class could register a callback that gets notified whenever a motor driver has new data.

dlech:
We might just have to let this be good enough for now :wink:

If you want to squash, I'll merge this as-is.

WasabiFan:
> If you want to squash, I'll merge this as-is.

Why have him squash instead of using GitHub's squash button?

dlech:
Because @kortschak wants to update the commit message as well.

WasabiFan:
> Because @kortschak wants to update the commit message as well.

You can do that with the web interface too :stuck_out_tongue_winking_eye: But I think I see your point.

kortschak:
Yes, but you can't selectively squash.

kortschak:
It's ready. Thank you for your patience and help throughout.

kortschak:
Hang on - incorrect fx link.

dlech:
Thanks!

PR #55, opened 2016-10-13
doc: note that durations must not be negative

kortschak:
Fixes ev3dev/ev3dev#744.

@dlech Please take a look.

### DIFF
--- a/motors/dc_motor_class.c
+++ b/motors/dc_motor_class.c
@@ -112,7 +112,7 @@
  * .
  * `time_sp`
  * : (read/write) Sets the time setpoint used with the `run-timed` command.
- *   Units are in milliseconds.
+ *   Units are in milliseconds. Values must not be negative.
  * .
  * [supported devices]: /docs/motors/#supported-devices
  */
--- a/motors/servo_motor_class.c
+++ b/motors/servo_motor_class.c
@@ -53,6 +53,7 @@
  *   tells the servo to drive to the maximum (clockwise) position_sp. Default
  *   value is 2400. Valid values are 2300 to 2700. You must write to the
  *   `position_sp` attribute for changes to this attribute to take effect.
+ *   Values must not be negative.
  * .
  * `mid_pulse_sp`
  * : (read/write) Used to set the pulse size in milliseconds for the signal that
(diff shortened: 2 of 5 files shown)


motors/dc_motor_class.c
```diff
@@ -112,7 +112,7 @@
  * .
  * `time_sp`
  * : (read/write) Sets the time setpoint used with the `run-timed` command.
- *   Units are in milliseconds.
+ *   Units are in milliseconds. Values must not be negative.
```
dlech: does writing a negative number to this attribute return `-EINVAL`? If not, it probably should.

dlech:
I guess it is just intuitive to me that you would never have a negative time value. It doesn't seem necessary to me to explicit state that.

motors/dc_motor_class.c
kortschak: I dont have a dc motor lead to test that with.

kortschak:
In a variety of contexts, negative duration values are used to indicate no-limit. These additions deny that here.

motors/dc_motor_class.c
kortschak: No worries. In many cases I have checked. I will do an exhaustive check tonight (currently apt upgrading and then going to work).

kortschak:
I've done the checks I can for whether the attributes do indeed baulk at negative values.

In all cases I tried `echo -1 > <attr>`. The checked attributes printed `write error: Invalid argument`.

dc-motor:
- [x] `time_sp`

servo-motor:
- [ ] `time_sp` (could not check cannot set out port mode to servo-motor)

tacho-motor:
- [x] `ramp_up_sp`
- [x] `ramp_down_sp`
- [x] `time_sp`

lego-sensor:
- [ ] `poll_ms` (not able to check - no available device with modifiable poll rate)

nxt-i2c:
- [x] `default_poll_ms` (done as root)

dlech:
> could not check - cannot set out port mode to servo-motor

The servo motor class doesn't have `time_sp` but the other time based setpoints have range checking. [example](https://github.com/ev3dev/lego-linux-drivers/blob/master/motors/servo_motor_class.c#L257).

> not able to check - no available device with modifiable poll rate

'poll_ms` is [parsing unsigned ints](https://github.com/ev3dev/lego-linux-drivers/blob/5ccdd2dc17f097d7a766ecad18b54d5372fbf88c/sensors/lego_sensor_class.c#L515), so we should be good here.

dlech:
So, LGTM. Will you update the commit message with a reference to the issue?

kortschak:
Yes, I'll do that.

kortschak:
>The servo motor class doesn't have `time_sp`

Ugh, I can't even read my own commits.

dlech:
Thanks!

PR #56, opened 2016-10-21
sensors: add Fatcatlab sensors

dlech:
This adds sensor definitions for Fatcatlab sensors. These are all
EV3/UART type sensors.

I don't have the Gesture Sensor, so it is missing from the list.

### DIFF
--- a/sensors/ev3_uart_sensor_defs.c
+++ b/sensors/ev3_uart_sensor_defs.c
@@ -1,7 +1,7 @@
 /*
  * LEGO MINDSTORMS EV3 UART Sensor driver
  *
- * Copyright (C) 2013-2015 David Lechner <david@lechnology.com>
+ * Copyright (C) 2013-2016 David Lechner <david@lechnology.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 as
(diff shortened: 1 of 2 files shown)


dlech:
@kortschak, do you happen to have the Gesture Sensor?

kortschak:
I do. what would you like me to try?

sensors/ev3_uart_sensor_defs.c
```diff
@@ -519,5 +519,259 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
+			[2] = {
+				/**
+				 * @description: Humidity
+				 * @value0: Humidity (0 to 1000)
+				 * @units_description: percent relative humidity
+				 */
+				.name		= "HUMIDITY",
+				.data_sets	= 1,
+				.data_type	= LEGO_SENSOR_DATA_S16,
+				.units		= "%RH",
```
kortschak: Does this `%` need to be `%%`?

dlech: Nope. This isn't going to be the `fmt` argument of any `printf()`.

dlech:
I don't have that sensor, so to make an entry for this sensor I need to know the `driver_name` and `modes` plus `bin_data_format`, `decimals`, `num_values` and `units` for each mode.

Also, the docs for the gesture sensor say something about resetting the value. I'm hoping that the resetting is being done in the EV3-G software and that there is not a special command for this sensor.

How do the values behave, for example, if you `watch -n 0.1 cat value0`? How long does the value of a gesture persist? Or does it get stuck on a value and never go back to 0?

kortschak:
```
$ cat driver_name 
ev3-uart-72
$ cat modes
GESTURE PROXIMITY RGB-RAW CLEAR
$ for m in $(cat modes); do echo $m >mode; cat decimals; done
0
0
0
0
$ cat num_values 
1
$ for m in $(cat modes); do echo $m >mode; cat num_values; done
1
1
3
1
$ for m in $(cat modes); do echo $m >mode; cat units; done




```

For `GESTURE`, the value persists until another gesture is made. For `PROXIMITY`, the value always reflects the proximity state. The `CLEAR` mode always returns 1, but changing mode from `GESTURE` -> `CLEAR` -> `GESTURE` takes the value back to zero, so that could be what they mean.

dlech:
Thanks. I've added the gesture sensor to the PR.

FYI, in the next kernel release, these sensors will return a meaningful name for `driver_name` instead of the generic type id (e.g. `fcl-gesture` instead of `ev3-uart-72`). And these sensors will show up on our website.

sensors/ev3_uart_sensor_defs.c
```diff
@@ -519,5 +519,327 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
+				 * @description: Magnetometer
+				 * @value0: X-axis magnetic flux density (-4800 to 4800)
+				 * @value1: Y-axis magnetic flux density (-4800 to 4800)
+				 * @value2: Z-axis magnetic flux density (-4800 to 4800)
+				 * @units_description: microteslas
+				 */
+				.name		= "MAGNET",
+				.data_sets	= 3,
+				.data_type	= LEGO_SENSOR_DATA_S32,
+				.units		= "uT",
```
kortschak: `u` is used as a substitute for `µ`. Is there a reason the unicode symbol cannot be used?

dlech:
Thanks for having a look at this.

sensors/ev3_uart_sensor_defs.c
kortschak: Thanks.

PR #57, opened 2016-10-29
Wait 100ms before setting the stalled flag

ddemidov:
See rhempel/ev3dev-lang-python#241.

This is more of a question than an actual PR. Would this be the correct way to delay setting the `stalled` state?

### DIFF
--- a/ev3/legoev3_motor.c
+++ b/ev3/legoev3_motor.c
@@ -642,13 +642,17 @@ static void calculate_speed(struct legoev3_motor_data *ev3_tm)
 		ev3_tm->speed = 0;
 		ev3_tm->stalled = 0;
 
-		if (ev3_tm->stalling) {
-			if (TACHO_MOTOR_STALLED_MS < ktime_to_ms(ktime_sub(ktime_get(), ev3_tm->stalling_since)))
-				ev3_tm->stalled = 1;
-		} else {
-			ev3_tm->stalling = 1;
-			ev3_tm->stalling_since = ktime_get();
-		}
+                if (ev3_tm->state == STATE_RUNNING) {
+                        if (ev3_tm->stalling) {
+                                if (TACHO_MOTOR_STALLED_MS < ktime_to_ms(ktime_sub(ktime_get(), ev3_tm->stalling_since)))
+                                        ev3_tm->stalled = 1;
+                        } else {
+                                ev3_tm->stalling = 1;
+                                ev3_tm->stalling_since = ktime_get();
+                        }
+                } else {
+                        ev3_tm->stalling = 0;
+                }
 	}
 
 	else if (TACHO_MOTOR_POLL_MS < ktime_to_ms(ktime_sub(ktime_get(), ev3_tm->tacho_samples[diff_idx]))) {
--- a/ev3/legoev3_motor.c
+++ b/ev3/legoev3_motor.c
@@ -643,7 +643,7 @@ static void calculate_speed(struct legoev3_motor_data *ev3_tm)
 		ev3_tm->stalled = 0;
 
 		if (ev3_tm->stalling) {
-			if ((TACHO_MOTOR_STALLED_MS * USEC_PER_SEC) < ktime_to_us(ktime_sub(ktime_get(), ev3_tm->stalling_since)))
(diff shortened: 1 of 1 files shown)


ev3/legoev3_motor.c
```diff
@@ -628,13 +633,22 @@ static void calculate_speed(struct legoev3_motor_data *ev3_tm)
...
 
 	} else if (ev3_tm->max_us_per_sample < ktime_to_us(ktime_sub(ktime_get(), ev3_tm->tacho_samples[diff_idx]))) {
 
 		ev3_tm->dir_chg_samples = 0;
 		ev3_tm->speed = 0;
-		ev3_tm->stalled = 1;
+		ev3_tm->stalled = 0;
+
+		if (ev3_tm->stalling) {
+			if ((TACHO_MOTOR_STALLED_MS * USEC_PER_SEC) < ktime_to_us(ktime_sub(ktime_get(), ev3_tm->stalling_since)))
```
dlech: Why not use `ktime_to_ms()`?

ddemidov:
No, not yet. Are instructions at https://github.com/ev3dev/ev3dev-buildscripts up to date?

dlech:
yes, they should be up-to-date.

ev3/legoev3_motor.c
ddemidov: Because I did not know it exists, and you use the same pattern a [couple of lines below](https://github.com/ddemidov/lego-linux-drivers/blob/3ef55b23c6a78d83b20fcdd642506e8612b6d5f9/ev3/legoev3_motor.c#L654) :)

ddemidov: Updated both instances in 65a12c8

ddemidov:
I'll compile, test, and report back here, but this may take some time (a couple of days).

ddemidov:
I've compiled and tested the changes, but I am getting strange results. This is the script I used for the tests:
```py
#!/usr/bin/env python3

import ev3dev.ev3 as ev3
import select
import time

m = ev3.LargeMotor()
_ = m.state

p = select.poll()
p.register(m._state, select.POLLPRI)

tic = time.time()
m.run_timed(speed_sp=400, time_sp=3000)

while True:
    p.poll()
    print('{}, {}'.format(time.time() - tic, m.state))
```

On the vanilla kernel I get the following output:
```
0.012087821960449219, ['running', 'stalled']
0.024071693420410156, ['running']
3.012042999267578, []
```
and the changes proposed here are supposed to get rid of the initial 'stalled' state. I am still getting the same output on the freshly compiled kernel though:
```
$ uname -a
Linux ev3dev 4.4.24-16-ev3dev-ev3stall-100-ev3+ #4 PREEMPT Sun Oct 30 14:55:18 MSK 2016 armv5tejl GNU/Linux
$ ./test.py
0.020143508911132812, ['running', 'stalled']
0.055086612701416016, ['running']
3.028170585632324, []
$ ./test.py
0.012041807174682617, ['running', 'stalled']
0.021996021270751953, ['running']
1.2285277843475342, ['running', 'overloaded']
1.9219954013824463, ['running', 'overloaded', 'stalled']
2.5086421966552734, ['running', 'overloaded']
2.5420291423797607, ['running']
2.550187349319458, ['running']
3.0200283527374268, []
```
(during the second run I grabbed the wheel to make it stall). I also tried increasing the [TACHO_MOTOR_STALLED_MS](https://github.com/ev3dev/lego-linux-drivers/pull/57/files#diff-e18526d541598fb5673911ef3d67ea05R65) constant up to 500, but still getting the initial 'stalled' in the output.

ddemidov:
Even increasing TACHO_MOTOR_STALLED_MS up to 1000 does not help. But I do see the pause between my grabbing the wheel and the appearance of 'stalled' in the output, so I am reasonably sure I am using the correct kernel and lego drivers.

ddemidov:
d1dcb50 does the trick for me. What it does is it only sets stalling flag and timer when motor is in `running` state. The test output:
```
$ uname -a
Linux ev3dev 4.4.24-16-ev3dev-ev3-stall-100-ev3+ #7 PREEMPT Sun Oct 30 16:47:24 MSK 2016 armv5tejl GNU/Linux

$ ./test.py 
0.023036479949951172, ['running']
3.025434732437134, []
```

dlech:
Awesome! Thanks!

PR #58, opened 2017-01-16
BrickPi workaround to make ms-absolute-imu work before i2c adapter driver

bmegli:
This is ugly hack and only for ms_absolute_imu. For mi-xg1300l it's far worse than that.

I don't really expect that @dlech merges it but this is the minimum code to make it work.

It relies on two facts:
- ms_imu uses only send_cmd_post_cb from ops
- in this function it only uses nxt_i2c_sensor_data.lego_sensor_device.mode_info

So the workaround creates mockup nxt_i2c_sensor_data with only this field filled. 

The bad thing is that everytime send_cmd is called a rather big mockup nxt_i2c_sensor_data structure is declared.

I have no means to test it (no ms_absolute_imu).

I am going to start working on i2c adapter driver in the free time (if nobody is working on it).

### DIFF
--- a/brickpi/brickpi_i2c_sensor.c
+++ b/brickpi/brickpi_i2c_sensor.c
@@ -137,7 +137,7 @@ static int brickpi_i2c_sensor_probe(struct lego_device *ldev)
 
 	sensor_info = &nxt_i2c_sensor_defs[ldev->entry_id->driver_data];
 
-	/* below temporary workaround for ms_absolute_imu (1 o 4) */
+	/* below temporary workaround for ms_absolute_imu (1 of 4) */
 	if (sensor_info->ops &&
 		ldev->entry_id->driver_data != MS_ABSOLUTE_IMU) {
 		dev_err(&ldev->dev, "The '%s' driver requires special operations"
--- a/brickpi/brickpi_i2c_sensor.c
+++ b/brickpi/brickpi_i2c_sensor.c
@@ -95,27 +95,26 @@ static int brickpi_i2c_sensor_send_command(void *context, u8 mode)
 	const struct nxt_i2c_sensor_cmd_info *i2c_cmd_info = data->info->i2c_cmd_info;
 	const struct nxt_i2c_sensor_mode_info *i2c_mode_info = data->info->i2c_mode_info;
 	int size = lego_sensor_get_raw_data_size(mode_info);
-	
-	int err; //temporary workaround for ms-absolute-imu (2 of 5)
-	struct nxt_i2c_sensor_data workaround_sensor; //workaround (3 of 5)
+
+	int err; /* temporary workaround for ms-absolute-imu (2 of 5) */
+	struct nxt_i2c_sensor_data workaround_sensor; /*workaround (3 of 5) */
 
 	/* set mode function also works for sending command */
 	err = brickpi_in_port_set_i2c_mode(data->ldev,
 					    i2c_cmd_info[mode].cmd_reg,
 					    i2c_cmd_info[mode].cmd_data,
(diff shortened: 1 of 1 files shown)


brickpi/brickpi_i2c_sensor.c
```diff
@@ -96,12 +96,27 @@ static int brickpi_i2c_sensor_send_command(void *context, u8 mode)
...
 					    i2c_cmd_info[mode].cmd_data,
 					    i2c_mode_info[mode].read_data_reg,
 					    size);
+
+	/* temporary workaround for ms-absolute-imu (4 of 5) */
+	if (err)
+		return err;
+
+	/* temporary workaround for ms-absolute-imu (5 of 5) */
+	if (data->info->ops && data->info->ops->send_cmd_post_cb) {
```
dlech: Do we not need `ldev->entry_id->driver_data != MS_ABSOLUTE_IMU` in the if statement here?

brickpi/brickpi_i2c_sensor.c
```diff
@@ -96,12 +96,27 @@ static int brickpi_i2c_sensor_send_command(void *context, u8 mode)
 	const struct nxt_i2c_sensor_mode_info *i2c_mode_info = data->info->i2c_mode_info;
 	int size = lego_sensor_get_raw_data_size(mode_info);
 
+	int err; /* temporary workaround for ms-absolute-imu (2 of 5) */
+	struct nxt_i2c_sensor_data workaround_sensor; /*workaround (3 of 5) */
```
dlech: This struct declaration can be moved inside of the if block.

bmegli: agreed

brickpi/brickpi_i2c_sensor.c
bmegli: any other sensor with non NULL ops wouldn't get through probe line:

``` 
if (sensor_info->ops && ldev->entry_id->driver_data != MS_ABSOLUTE_IMU) {
```

 but I agree that extra safety here would not hurt

bmegli:
Updated with safety check for MS_ABSOLUTE_IMU workaround and struct declaration in if.

Side note:

It's probably not strictly necessary but I have seen this pattern in some places in your code and it's not here.
Don't you want `dev_set_drvdata(&ldev->dev, NULL);` in line [185 and a half](https://github.com/ev3dev/lego-linux-drivers/blob/5ccdd2dc17f097d7a766ecad18b54d5372fbf88c/brickpi/brickpi_i2c_sensor.c#L185)?
(set driver data to NULL in case probe fails)

e.g.
```
err_register_lego_sensor:
        dev_set_drvdata(&ldev->dev, NULL);
	kfree(data->sensor.mode_info);
err_kalloc_mode_info:
	kfree(data);

	return err;
```

dlech:
> Don't you want dev_set_drvdata(&ldev->dev, NULL); in line 185 and a half?

I like having that just for neatness, but it is not really necessary and most mainline drivers omit it.

brickpi/brickpi_i2c_sensor.c
```diff
@@ -121,7 +137,9 @@ static int brickpi_i2c_sensor_probe(struct lego_device *ldev)
 
 	sensor_info = &nxt_i2c_sensor_defs[ldev->entry_id->driver_data];
 
-	if (sensor_info->ops) {
+	/* below temporary workaround for ms_absolute_imu (1 o 4) */
```
dlech: should be `1 of 4`

bmegli:
fixed "o" -> "of" typo in comment

dlech:
Tested working, thanks!

PR #59, opened 2017-01-25
increase EV3_UART_BUFFER_SIZE to 1024 bytes

bmegli:
At current 256 bytes buffer there are occasional overruns leading to UART sensor reconnections.

This is especially evident when changing buffer size to 128 bytes where under high load overruns are notorious.

1024 bytes buffer was tested in high load  scenario for 1 hour without overruns.

Note:

Also when reconnection happens surrounded with multiple overruns there seems to be some extremely rare edge case (race?) when sensor changes mode after reconnection.  This pull will likely hide this edge case problem. It will just not be hit anymore (no reconnections). It's hard to reproduce even with plenty of overruns. This is another thing to consider.

### DIFF
--- a/sensors/ev3_uart_sensor_ld.c
+++ b/sensors/ev3_uart_sensor_ld.c
@@ -76,7 +76,7 @@
 #endif
 
 /* EV3_UART_BUFFER_SIZE must be power of 2 for circ_buf macros */
-#define EV3_UART_BUFFER_SIZE		256
+#define EV3_UART_BUFFER_SIZE		1024
 #define EV3_UART_MAX_DATA_SIZE		32
 #define EV3_UART_MAX_MESSAGE_SIZE	(EV3_UART_MAX_DATA_SIZE + 2)
 
--- a/sensors/ev3_uart_sensor_ld.c
+++ b/sensors/ev3_uart_sensor_ld.c
@@ -76,7 +76,7 @@
 #endif
 
 /* EV3_UART_BUFFER_SIZE must be power of 2 for circ_buf macros */
-#define EV3_UART_BUFFER_SIZE		128
+#define EV3_UART_BUFFER_SIZE		1024
 #define EV3_UART_MAX_DATA_SIZE		32
 #define EV3_UART_MAX_MESSAGE_SIZE	(EV3_UART_MAX_DATA_SIZE + 2)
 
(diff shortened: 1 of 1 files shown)


dlech:
> Also when reconnection happens surrounded with multiple overruns there seems to be some extremely rare edge case (race?) when sensor changes mode after reconnection. This pull will likely hide this edge case problem. It will just not be hit anymore (no reconnections). It's hard to reproduce even with plenty of overruns. This is another thing to consider.

I'm not exactly sure what you mean here. If a sensor has to reconnect, the sensor is always in mode 0 when it connects. The driver remembers the mode the user selected and tries restore that mode after reconnection. Were you seeing a change back to mode 0 sometimes?

dlech:
Thanks for testing this!

bmegli:
@dlech wrote:

> I'm not exactly sure what you mean here. If a sensor has to reconnect, the sensor is always in mode 0 when it connects. The driver remembers the mode the user selected and tries restore that mode after reconnection. Were you seeing a change back to mode 0 sometimes?

Yes, this happened twice during testing. After reconnection the sensor ended in mode 0 (COL-REFLECT) even though it was in different (RGB-RAW) before reconnection. This is rare behaviour. It almost always restores last mode.

Also when sensor reconnects, probably as it receives mode data, it reports that it is in modes one by one, like in logs from the issue (those were not my logs, but mine were similiar when reconnecting):

``` bash
while [ 1 ]; do  cat /sys/class/lego-sensor/sensor0/mode; done
RGB-RAW Thu Dec  8 23:05:56 UTC 2016
RGB-RAW Thu Dec  8 23:05:57 UTC 2016
RGB-RAW Thu Dec  8 23:05:57 UTC 2016
RGB-RAW Thu Dec  8 23:05:58 UTC 2016
                  Thu Dec  8 23:05:58 UTC 2016
COL-REFLECT Thu Dec  8 23:05:58 UTC 2016
RGB-RAW Thu Dec  8 23:05:59 UTC 2016
RGB-RAW Thu Dec  8 23:05:59 UTC 2016
RGB-RAW Thu Dec  8 23:06:00 UTC 2016
COL-CAL Thu Dec  8 23:06:00 UTC 2016
REF-RAW Thu Dec  8 23:06:01 UTC 2016
COL-AMBIENT Thu Dec  8 23:06:01 UTC 2016
RGB-RAW Thu Dec  8 23:06:02 UTC 2016
RGB-RAW Thu Dec  8 23:06:02 UTC 2016
RGB-RAW Thu Dec  8 23:06:02 UTC 2016
```

dlech:
Ah, I see. I looks like we are using the publicly readable mode variable when enumerating modes. We should be using a private variable here.

PR #60, opened 2017-02-06
Fix footnotes for EV3 IR sensor IR-SEEK mode

ddemidov:
See ev3dev/ev3dev#834, rhempel/ev3dev-lang-python#291


dlech:
Looks good. Thanks!

PR #61, opened 2017-03-16
Correct file path in i2c documentation

seth10:
I'm not sure if this is correct, but I don't see a `modprobe` directory in `/etc` on my brick, but I do see a `modprobe.d`, which I'm guessing is what this meant.


dlech:
Thanks!

PR #62, opened 2017-03-16
Remove trailing whitespace Travis CI was complaining about

seth10:


seth10:
The last error is `ERROR: do not set execute permissions for source files` for `FILE: Documentation/sphinx/kernel-doc`. I tried a `chmod -x` on it through Cygwin, but that must not have made it in the commit. Could someone on a native UNIX platform try?

seth10:
Never mind, got it! Thanks @ethomson ^~^ ([SO](http://stackoverflow.com/a/21694391/5279083))

seth10:
Now the result is `total: 0 errors, 239 warnings, 13806 lines checked`. Although there are 0 errors, it still fails the checks, I guess warnings are more significant to it than I realized.

dlech:
Hmm.. travis is only supposed to be checking the lines you actually changed. I suspect this has to do with  the fact that I changed the default branch name recently. I'll take the whitespace fixes, but `kernel-doc` is supposed to be executable and documentation builds will break if it is not executable.

dlech:
And FWIW, travis only knows pass or fail based on the return code, so warnings and errors are all the same to travis. We don't care about making travis green on this repository, but the automatic checks are there just to point out obvious problems. I missed all of the whitespace problems because I made such a huge commit with the recent documentation changes.

evb/evb_input.c
```diff
@@ -25,25 +25,25 @@
...
+ *
  *     * - Back button press
  *       - ``KEY_BACKSPACE`` (14)
- * 
+ *
  *     * - Joystick button press
  *       - ``KEY_ENTER`` (28)
- * 
+ *
  *     * - Joystic up
```
dlech: Look like we can fix the spelling of joystic here too. And several repeats below.

dlech (changes requested):

seth10:
Okay, thanks for the feedback! When I saw my other PR had a big "All checks have failed," my thought was to fix that. So yes certainly _don't_ merge the `kernel-doc` permission change.

> We don't care about making travis green on this repository

Got it, should I close this?

dlech:
If you can drop the second commit, I will merge the first one. I also noticed a spelling error near the whitespace changes that can be fixed too.

evb/evb_input.c
seth10: Good call! Corrected in cd47d22305c19c51625931d216c0bcafcf8021a5

dlech (approved):

dlech:
Awesome, thanks!

PR #67, opened 2017-04-12
Support for BrickPi3 Firmware V1.4.2

sanjayseshan:


brickpi3/brickpi3.h
```diff
@@ -37,87 +37,52 @@ enum brickpi3_output_port {
...
+	BRICKPI3_MSG_GET_SENSOR_2,
+	BRICKPI3_MSG_GET_SENSOR_3,
+	BRICKPI3_MSG_GET_SENSOR_4,
 #define	BRICKPI3_MSG_I2C_TRANSACT \
 	BRICKPI3_MSG_I2C_TRANSACT_1
-	BRICKPI3_MSG_I2C_TRANSACT_1 = 64,
+	BRICKPI3_MSG_I2C_TRANSACT_1,
 	BRICKPI3_MSG_I2C_TRANSACT_2,
-	BRICKPI3_MSG_I2C_TRANSACT_3,
+       	BRICKPI3_MSG_I2C_TRANSACT_3,
```
dlech: Don't need the extra spaces here before the tab

brickpi3/brickpi3.h
```diff
@@ -174,8 +140,12 @@ enum brickpi3_i2c_flags {
 	BRICKPI3_I2C_SAME	= 0x04,
 };
 
-#define BRICKPI3_I2C_MAX_WRITE_SIZE	16
-#define BRICKPI3_I2C_MAX_READ_SIZE	14
+enum brickpi3_motor_status {
+    BRICKPI3_MOTOR_STATUS_LOW_VOLTAGE_FLOAT = 0x01,
+};
+
+#define BRICKPI3_I2C_MAX_SET_SIZE	16
```
dlech: I would rather not rename this constant. "get" and "set" are more high-level terms and and "read" and "write" are more low-level, generally speaking. I2C is more of a low level thing, so I would like to keep it as `BRICKPI3_I2C_MAX_WRITE_SIZE`

brickpi3/brickpi3.h
```diff
@@ -174,8 +140,12 @@ enum brickpi3_i2c_flags {
...
+#define BRICKPI3_I2C_MAX_GET_SIZE	14
```
dlech: Same with this one, I would like to keep it as `BRICKPI3_I2C_MAX_READ_SIZE`

brickpi3/brickpi3_i2c.c
```diff
@@ -43,7 +43,7 @@ static int brickpi3_i2c_master_xfer(struct i2c_adapter *adap,
 
 	/*
 	 * Hopefully i2c core is doing lots of checking for us based on
-	 * I2C_AQ_COMB_WRITE_THEN_READ. We only support one or two messages,
+	 * I2C_AQ_COMB_SET_THEN_READ. We only support one or two messages,
```
dlech: This should not be renamed. It comes from elsewhere in the Linux kernel.

brickpi3/brickpi3_ports_out.c
```diff
@@ -246,8 +246,11 @@ static int brickpi3_out_port_run_to_pos(void *context, int pos, int speed,
 {
 	struct brickpi3_out_port *data = context;
 	int ret;
-
-	/* FIXME: speed and stop action are ignored */
+	
+	ret = brickpi3_set_motor_limits(data->bp, data->index, 0, speed);
+	if (ret < 0)
+	  return ret;
```
dlech: Use tab instead of spaces please

brickpi3/brickpi3_spi.c
```diff
@@ -12,7 +12,7 @@
 
  * This program is distributed "as is" WITHOUT ANY WARRANTY of any
  * kind, whether express or implied; without even the implied warranty
- * of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * of MERCHANTABILITY r FITNESS FOR A PARTICULAR PURPOSE.  See the
```
dlech: Not sure why this is changed

brickpi3/brickpi3_spi.c
```diff
@@ -24,13 +24,13 @@
 
 #include "brickpi3.h"
 
-#define BRICKPI3_REQUIRED_FIRMWARE_VERSION	1002000 /* 1.2.0 */
+#define BRICKPI3_REQUIRED_FIRMWARE_VERSION	1004000 /* 1.2.0 */
```
dlech: the comment needs to be changed to match the value.

brickpi3/brickpi3_spi.c
```diff
@@ -24,13 +24,13 @@
...
 #define BRICKPI3_HEADER_SIZE		4
 #define BRICKPI3_ID_MSG_SIZE		16
 #define BRICKPI3_STRING_MSG_SIZE	20
 #define BRICKPI3_MAX_MSG_SIZE (BRICKPI3_HEADER_SIZE + BRICKPI3_STRING_MSG_SIZE)
 
-#define BRICKPI3_READ_FAILED(b)	((b)[3] != 0xA5)
+#define BRICKPI3_GET_FAILED(b)	((b)[3] != 0xA5)
```
dlech: I would rather not rename this macro either. It is more of a low-level operation.

brickpi3/brickpi3_spi.c
```diff
@@ -138,6 +138,69 @@ int brickpi3_write_u16(struct brickpi3 *bp, enum brickpi3_message msg,
...
+	bp->xfer.len = 5;
+
+	ret = spi_sync(bp->spi, &bp->msg);
+
+	mutex_unlock(&bp->xfer_lock);
+
+	return ret;
+}
+
+int brickpi3_set_motor_limits(struct brickpi3 *bp, enum brickpi3_output_port port, u8 duty_cycle_sp,
```
dlech: I would like to move this function after `brickpi3_i2c_transact()` This way, it is in the same order as it is listed in `brickpi3.h`. And it keeps things a bit more organized since we have the generic helper functions all grouped together here and then all of the other specialized functions later.

brickpi3/brickpi3_spi.c
```diff
@@ -243,6 +306,32 @@ int brickpi3_write_u32(struct brickpi3 *bp, enum brickpi3_message msg,
...
+	bp->buf[6] = value2 & 0xff;
+	bp->xfer.len = 7;
+
+	ret = spi_sync(bp->spi, &bp->msg);
+
+	mutex_unlock(&bp->xfer_lock);
+
+	return ret;
+}
+
```
dlech, line 333: Just one blank line between functions please

brickpi3/brickpi3_spi.c
```diff
@@ -138,6 +138,69 @@ int brickpi3_write_u16(struct brickpi3 *bp, enum brickpi3_message msg,
...
+	bp->buf[3] = value2;
+	bp->xfer.len = 4;
+
+	ret = spi_sync(bp->spi, &bp->msg);
+
+	mutex_unlock(&bp->xfer_lock);
+
+	return ret;
+}
+
```
dlech, line 204: just one blank line between functions please

brickpi3/brickpi3_spi.c
```diff
@@ -404,16 +495,17 @@ int brickpi3_set_sensor_i2c(struct brickpi3 *bp,
 	return ret;
 }
 
+
```
dlech: No need to add a blank line here

brickpi3/brickpi3_spi.c
```diff
@@ -441,14 +533,14 @@ int brickpi3_i2c_transact(struct brickpi3 *bp, enum brickpi3_input_port port,
 	bp->buf[2] = addr << 1;
 	bp->buf[3] = read_size;
 	bp->buf[4] = write_size;
-	memcpy(&bp->buf[5], write_buf, write_size);
-	bp->xfer.len = 5 + write_size;
+       	memcpy(&bp->buf[5], write_buf, write_size);
```
dlech: No spaces before the tab here please

brickpi3/brickpi3_spi.c
```diff
@@ -441,14 +533,14 @@ int brickpi3_i2c_transact(struct brickpi3 *bp, enum brickpi3_input_port port,
...
+	bp->xfer.len = 6 + write_size;
```
dlech: I don't think we should be changing 5 => 6 here.

dlech (changes requested):
Looks pretty good. There are a couple of things I would rather not rename and some style issues to clean up, but other than that, I only found one potential issue.

brickpi3/brickpi3_spi.c
```diff
@@ -390,11 +481,12 @@ int brickpi3_set_sensor_i2c(struct brickpi3 *bp,
 	mutex_lock(&bp->xfer_lock);
 
 	bp->buf[0] = bp->spi->chip_select;
-	bp->buf[1] = BRICKPI3_MSG_SET_SENSOR_TYPE + port;
-	bp->buf[2] = BRICKPI3_SENSOR_TYPE_I2C;
-	bp->buf[3] = flags;
-	bp->buf[4] = speed;
-	bp->xfer.len = 5;
+	bp->buf[1] = BRICKPI3_MSG_SET_SENSOR_TYPE;
+	bp->buf[2] =  BIT(port);
```
dlech: there should only be one space before BIT

brickpi3/brickpi3_i2c.c
sanjayseshan: I changed this to match the python file

brickpi3/brickpi3_spi.c
sanjayseshan: will fix

brickpi3/brickpi3_spi.c
sanjayseshan: This was done to match the python file

brickpi3/brickpi3_i2c.c
dlech: This name comes from the Linux kernel and is never mentioned in brickpi3.py

brickpi3/brickpi3_spi.c
dlech: This macro is something I created for the ev3dev drivers and it is not mentioned in brickpi3.py (you can have look at the usage of 0xA5 in brickpi3.py to see what I mean)

dlech:
I found 3 places in brickpi3_ports_in.c where we are calling `brickpi3_write_u8()`. We missed these because it is a generic helper function instead of a specialized function. To fix this, I think we should make a new inline function in brickpi.h called `brickpi3_set_sensor_type()`.

dlech:
Merged in a528ec9 and 1f66e3e

PR #68, opened 2017-10-15
sensors: return empty list for sensors with no commands

kortschak:
@dlech Please take a look.

It looks to me like the `command` attribute was not behaving as described, but I could be reading it wrong. It says it returns `ENOTSUP` if the sensor doesn't have any commands, but the only error return I see is `EINVAL`. I've changed the documentation to match the behaviour, but I can do the reverse if you want. For reference, the behaviour looks the same as for the tacho motor class.

Fixes ev3dev/ev3dev#986.

### DIFF
--- a/sensors/lego_sensor_class.c
+++ b/sensors/lego_sensor_class.c
@@ -80,13 +80,12 @@
  *    * - ``command``
  *      - write-only
  *      - Sends a command to the sensor. See the individual sensor documentation
- *        for possible commands. Sensors that do not support commands will
- *        return ``-EOPNOTSUPP`` when writing to this attribute.
+ *        for possible commands.
  *
  *    * - ``commands``
  *      - read-only
  *      - Returns a space separated list of the valid commands for the
- *        sensor. Returns ``-EOPNOTSUPP`` if no commands are supported.
+ *        sensor.
  *
  *    * - ``direct``
  *      - read/write
@@ -324,10 +323,13 @@ static ssize_t commands_show(struct device *dev, struct device_attribute *attr,
 	int i;
 	unsigned count = 0;
 
+	if (sensor->num_commands == 0) {
+		buf[0] = '\n';
+		return 1;
+	}
+
 	for (i = 0; i < sensor->num_commands; i++)
 		count += sprintf(buf + count, "%s ", sensor->cmd_info[i].name);
-	if (count == 0)
-		return -EOPNOTSUPP;
 	buf[count - 1] = '\n';
 
 	return count;


sensors/lego_sensor_class.c
```diff
@@ -326,8 +325,7 @@ static ssize_t commands_show(struct device *dev, struct device_attribute *attr,
 
 	for (i = 0; i < sensor->num_commands; i++)
 		count += sprintf(buf + count, "%s ", sensor->cmd_info[i].name);
-	if (count == 0)
-		return -EOPNOTSUPP;
+
 	buf[count - 1] = '\n';
```
dlech: Have you tested this with a sensor with no commands? It seems to me that we would crash here if `count` == 0 because that would make the array index -1.

Maybe move the `if` statement before the `for` loop:

```
if (!count) {
        buf[0] = '\n';
        return 1;
}
```

dlech:
The idea behind `EOPNOTSUPP` on the `command` attribute was to differentiate between an invalid command on sensors that do support commands and a sensor that does not support commands at all. Thinking about it now though, I'm not sure that this distinction would be useful. So, I'm inclined to just leave it the way it is now and update the docs as you have suggested.

sensors/lego_sensor_class.c
kortschak: I thought the same thing, but the flow is identical to the case in [`tacho_motor_class.c`](https://github.com/ev3dev/lego-linux-drivers/blob/526851b7b7a2e90ff46febdc48a55735ef7d7bfd/motors/tacho_motor_class.c#L625-L643). Is this because there is never a case where `size` is zero there?

dlech: Correct. In the tacho-motor class, we can assume that there will never be an empty list, but for sensors, there can be an empty list for `commands`.

kortschak: No worries, I fix that tonight (I only just noticed your replies - I have been spammed by our HPC queue telling me of hundreds of node failures and everything else just got pushed down).

kortschak:
PTAL

sensors/lego_sensor_class.c
```diff
@@ -324,10 +323,14 @@ static ssize_t commands_show(struct device *dev, struct device_attribute *attr,
...
+	if (sensor->num_commands == 0) {
+		buf[0] = '\n';
+		return 1;
+	}
+
 	for (i = 0; i < sensor->num_commands; i++)
 		count += sprintf(buf + count, "%s ", sensor->cmd_info[i].name);
-	if (count == 0)
-		return -EOPNOTSUPP;
+
```
dlech: nit: no need to add a line here.

dlech:
Looks like it should work. Have you actually tested it?

sensors/lego_sensor_class.c
kortschak: Removed

kortschak:
I don't have a stretch release installed yet, but I just tried to build the stretch kernel and it fails for me in the lms2012 compatility module:

```
/home/daniel/Development/ev3dev/ev3-kernel/drivers/lego/lms2012/lms2012_compat.c: In function 'lms2012_compat_probe':
/home/daniel/Development/ev3dev/ev3-kernel/drivers/lego/lms2012/lms2012_compat.c:144:21: error: implicit declaration of function 'devm_pinctrl_get_select_default' [-Werror=implicit-function-declaration]
   lms->pinctrl[i] = devm_pinctrl_get_select_default(&in_port->dev);
                     ^
/home/daniel/Development/ev3dev/ev3-kernel/drivers/lego/lms2012/lms2012_compat.c:144:19: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
   lms->pinctrl[i] = devm_pinctrl_get_select_default(&in_port->dev);
                   ^
/home/daniel/Development/ev3dev/ev3-kernel/drivers/lego/lms2012/lms2012_compat.c:151:29: error: implicit declaration of function 'pinctrl_lookup_state' [-Werror=implicit-function-declaration]
   lms->pinctrl_default[i] = pinctrl_lookup_state(lms->pinctrl[i],
                             ^
/home/daniel/Development/ev3dev/ev3-kernel/drivers/lego/lms2012/lms2012_compat.c:151:27: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
   lms->pinctrl_default[i] = pinctrl_lookup_state(lms->pinctrl[i],
                           ^
/home/daniel/Development/ev3dev/ev3-kernel/drivers/lego/lms2012/lms2012_compat.c:153:23: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
   lms->pinctrl_i2c[i] = pinctrl_lookup_state(lms->pinctrl[i],
                       ^
  AS      arch/arm/lib/muldi3.o
  AS      arch/arm/lib/putuser.o
  AS      arch/arm/lib/setbit.o
  AS      arch/arm/lib/strchr.o
  AS      arch/arm/lib/strrchr.o
cc1: some warnings being treated as errors
/home/daniel/Development/ev3dev/ev3-kernel/scripts/Makefile.build:299: recipe for target 'drivers/lego/lms2012/lms2012_compat.o' failed
make[8]: *** [drivers/lego/lms2012/lms2012_compat.o] Error 1
/home/daniel/Development/ev3dev/ev3-kernel/scripts/Makefile.build:544: recipe for target 'drivers/lego/lms2012' failed
make[7]: *** [drivers/lego/lms2012] Error 2
/home/daniel/Development/ev3dev/ev3-kernel/scripts/Makefile.build:544: recipe for target 'drivers/lego' failed
make[6]: *** [drivers/lego] Error 2
/home/daniel/Development/ev3dev/ev3-kernel/Makefile:991: recipe for target 'drivers' failed
make[5]: *** [drivers] Error 2
make[5]: *** Waiting for unfinished jobs....
```

I will try to find some time to get a stretch card written over the next few days and then reattempt the kernel build.

dlech:
Could be an issue with `.config`. You can use `./defconfig merge` in the ev3dev build scripts (FYI, the build scripts repo also has separate jessie and stretch branches). Or you can just delete `.config` and it will create a new one automatically.

kortschak:
>FYI, the build scripts repo also has separate jessie and stretch branches

That may be a cause. Thanks.

kortschak:
Tested and working.

dlech:
Great, thanks!

PR #72, opened 2019-01-25
Changed the in #1210 described line to fic ENOMEM error with HT Sensor Multiplexer

w1ref1re:
I did like in https://github.com/ev3dev/ev3dev/issues/1210 described, and the `HT SMUX` is now detecting the Sensors. But the values are always `255`, they are not changing,


dlech:
Thanks for the pull request! I've found the other problem and fixed it too.

PR #73, opened 2019-06-03
add support for mindsensors TOF sensor

spookydonut:
this code is untested but based on my i2c probing of it should in theory work.


dlech:
Hi, sorry I totally missed this PR somehow :confused: 

Great job though! I went ahead and added an extra mode for the raw voltage and a command to power off the sensor so that we aren't missing out on any functionality.

Thanks!

PR #74, opened 2019-07-22
sensors/ev3_uart_sensor_defs: Update doc: IR-REM-A

laurensvalk:
Apparently, bit 3 toggles sometimes, which means there are two values for each button combination.

Also, 352 & 0x0F == 0, which breaks the example for blue up & red down. Instead, just check for 384.


laurensvalk:
I think the table is 79 characters in width now. Not sure about these length warnings.

dlech:
Thanks. (don't worry about the warnings for doc comment lines)

PR #75, opened 2019-09-29
Fixed array or angles for Absolute-IMU ACG

omartin2010:
as discussed


dlech:
This is not quite correct. I've done some testing with the sensor and a protractor and found that the old scaling was OK at 0, 90 and 180 but not at 45 and 135 degrees. I've fixed that in https://github.com/ev3dev/lego-linux-drivers/commit/5441613f1e20ff234d05c4862d84ef52671ac820.

This assumes that when an arrow on the sensor is pointing up, that is 0 degrees, when the arrow is pointing down, that is 180 degrees and when the arrow is horizontal, that is 90 degrees.

But I'm wondering if in practice it would make more  sense for up to be 90 degrees, horizontal to be 0 degrees and down to be -90 degrees.

dlech:
I went ahead and did the -90 to 90 in https://github.com/ev3dev/lego-linux-drivers/commit/ada9c0e96ff67143c3f4d3cb0d7f9fe796009b9f.

This is a breaking change and not fixing a bug, so I'm only going to do it for ev3dev-buster and not backport it to ev3dev-stretch.

omartin2010:
I think it does make more sense. Although the sensor's precision is so bad around +-90, that I built a complementary filter instead in Python that runs in a thread (1 thread/sensor) but it would be awesome if it would instead be provided in the driver :). But not sure there's a real use case for that to be there other than mine...

dlech:
Yeah, I've been noticing that at close to 90 degrees, sometimes the value of some axes will suddenly become negative while others won't get all the way to 90.

omartin2010:
or you get 83, then 90 and nothing in between. That's what has driven me to find alternative and learn about these filters. Pretty cool, and very precise at any angle, with 6 dof. Although I only use it around 1 axis, I get the data for all axis...

PR #77, opened 2019-11-16
sensors: fix HT_NXT_SUPER_PRO_SENSOR command list

keenanjohnson:
HT_NXT_SUPER_PRO_SENSOR only enumerated 5 out of it's 8 supported commands.

### DIFF
--- a/sensors/nxt_i2c_sensor_defs.c
+++ b/sensors/nxt_i2c_sensor_defs.c
@@ -1399,7 +1399,7 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 		.vendor_id	= "HiTechnc",
 		.product_id	= "SuperPro",
 		.pin1_state	= LEGO_PORT_GPIO_HIGH,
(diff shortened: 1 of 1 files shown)


keenanjohnson:
I'm not sure if there is anything else I need to update or if this five was put here intentionally, but it seems wrong to me.

[See the docs here for the list of all the commands as a double check.](http://docs.ev3dev.org/projects/lego-linux-drivers/en/ev3dev-stretch/sensor_data.html#hitechnic-nxt-superpro-prototype-board)

image

dlech:
Thanks!

For the record, we also noticed that there are some issues with the data conversion - e.g. AIN is not 16-bit little-endian, but rather the first byte is the 8 highest bits and the second byte is the two lowest bits. So there is more work to be done on this sensor to make it work properly.

keenanjohnson:
Yes! Are there issues for this project? Should I open one so we don't forget?

dlech:
https://github.com/ev3dev/ev3dev/issues

PR #78, opened 2020-03-04
Document less known EV3 color sensor modes

### LINKED ISSUE #1386: Meaning of raw values from the EV3 color sensor
**Is your feature request related to a problem? Please describe.**
No, it just stems from my curiosity.

**Describe the solution you'd like**
Adding documentation about the raw modes of the EV3 color sensor (in https://github.com/ev3dev/lego-linux-drivers/blob/4cddeee8d48f3fc01dff3be59906c9a5a2c1c635/sensors/ev3_uart_sensor_defs.c#L87)

**Describe alternatives you've considered**
N/A

**Additional context**
I have been playing with [c4ev3/EV3-API](https://github.com/c4ev3/EV3-API/) and there is a currently a new userspace sensor API in development. I wanted to know what the RGB-RAW and REF-RAW values mean and I think I found the answer.
REF-RAW: https://github.com/simonedegiacomi/EV3-API/commit/20a57f7b944a1e53477ebdfe61f96ebf2d6c66b7#r37442226
RGB-RAW: https://github.com/simonedegiacomi/EV3-API/commit/8562bd29da2ab430807741f7374ac60434eb793e#r37442413 (followup to REF-RAW)

However perhaps the documentation of the raw values should be put on its own page (if included with all the details), as it may be too much information for one sensor/mode.

I haven't proven if the values comes directly from the ADC by disassembling the *sensor* firmware, but I don't think there's lot the sensor could do with the values.

I also have an idea about what the `COL-CAL` mode does and why it behaves like it does. However, in this case, I have no proof of whether this is true. I think the mode is used in factory in order to provide calibration values for the `COL-REFLECT`, `COL-AMBIENT` and `COL-COLOR` modes to work correctly. Therefore, the sensor waits for extra data to come after the mode switch occurs and this makes the keepalive time out. I am also afraid a little to do experiments with sending random data to the sensor, as that could erase the existing data.

Full proof would be the full disassembly of the color sensor firmware. Some years ago, I have downloaded the flash/eeprom contents from the sensor. I currently do not have time to disassemble it, but I can send it via email if anyone wants to play with it.

Best regards,

Jakub Vanek

JakubVanek:
Fixes https://github.com/ev3dev/ev3dev/issues/1386
References:
* REF-RAW: https://github.com/simonedegiacomi/EV3-API/commit/20a57f7b944a1e53477ebdfe61f96ebf2d6c66b7#r37442226
* RGB-RAW: https://github.com/simonedegiacomi/EV3-API/commit/8562bd29da2ab430807741f7374ac60434eb793e#r37442413
* COL-CAL: https://github.com/ev3dev/ev3dev/issues/1386

### DIFF
--- a/sensors/ev3_uart_sensor_defs.c
+++ b/sensors/ev3_uart_sensor_defs.c
@@ -191,7 +191,7 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
 				.name		= "COL-CAL",
 				.data_sets	= 4,
 				.num_values	= 3,
-				.data_type	= LEGO_SENSOR_DATA_S32,
+				.data_type	= LEGO_SENSOR_DATA_U16,
 			},
 		},
 	},
--- a/sensors/ev3_uart_sensor_defs.c
+++ b/sensors/ev3_uart_sensor_defs.c
@@ -86,9 +86,35 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
 			},
 			[3] = {
 				/**
-				 * @description: Raw Reflected - sets LED color to red
-				 * @value0: ??? (0 to 1020???)
-				 * @value1: ??? (0 to 1020???)
+				 * .. [#lego-ev3-color-mode3-values] This mode
+				 *    provides direct access to the measurements
+				 *    done by the ADC inside the sensor. It
+				 *    works by continuously flashing the red LED
+				 *    on and off and measuring the current
+				 *    flowing through a bypass capacitor. It is
+				 *    not possible to measure true ambient light
+				 *    this way, but the LED off measurement
+				 *    works as a reference for the LED on
+				 *    measurement. Higher photodiode
+				 *    illumination will cause lower ADC
+				 *    measurements, so to obtain reflected light
+				 *    intensity similar to ``COL-REFLECT``, it
+				 *    is necessary to do
+				 *    ``reflect = value1 - value0``.
+				 *
+				 * .. [#lego-ev3-color-mode3-value1] While
+				 *    this value cannot be used to measure
+				 *    still ambient light, it can be used to
+				 *    measure *changes* in external
+				 *    illumination. When it increases, the
+				 *    this value goes below its baseline and
+				 *    vice versa.
+				 *
+				 * @description: Raw Reflected Light - sets LED color to red
+				 * @value0: ADC reading for LED on (reflection) (0 to 1023)
+				 * @value0_footnote: [#lego-ev3-color-mode3-values]_
+				 * @value1: ADC reading for LED off (reference) (0 to 1023)
+				 * @value1_footnote: [#lego-ev3-color-mode3-values]_ [#lego-ev3-color-mode3-value1]_
 				 */
 				.name		= "REF-RAW",
 				.data_sets	= 2,
--- a/sensors/ev3_uart_sensor_defs.c
+++ b/sensors/ev3_uart_sensor_defs.c
@@ -122,11 +122,33 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
 			},
 			[4] = {
 				/**
+				 * .. [#lego-ev3-color-mode4-value012] These
+				 *    values represent the difference between
+				 *    the raw sensor ADC measurements when the
+				 *    LEDs are off and when they are on. They
+				 *    work similar to the ``COL-REFLECT`` values
+				 *    but they are not precalibrated.
+				 *
+				 * .. [#lego-ev3-color-mode4-value3] This value
+				 *    is similar to the ``REF-RAW`` ambient
+				 *    ``value1``. It is the value measured by
+				 *    the ADC inside the sensor when all LEDs
+				 *    are off, so it works as a reference point
+				 *    for the first three values. However here
+				 *    the ``value0``, ``value1`` and ``value2``
+				 *    are sent as relative to this value by the
+				 *    sensor automatically, so it is not
+				 *    necessary to use this value at all.
+				 *
 				 * @description: Raw Color Components - sets LED color to white (all LEDs rapidly cycling)
-				 * @value0: Red (0 to 1020???)
-				 * @value1: Green (0 to 1020???)
-				 * @value2: Blue (0 to 1020???)
-				 * @value3: Ambient? (0 to 1020???)
+				 * @value0: Reflected red light intensity (0 to 1023)
+				 * @value0_footnote: [#lego-ev3-color-mode4-value012]_
+				 * @value1: Reflected green light intensity (0 to 1023)
+				 * @value1_footnote: [#lego-ev3-color-mode4-value012]_
+				 * @value2: Reflected blue light intensity (0 to 1023)
+				 * @value2_footnote: [#lego-ev3-color-mode4-value012]_
+				 * @value3: ADC reading for LEDs off (reference) (0 to 1023)
+				 * @value3_footnote: [#lego-ev3-color-mode4-value3]_
 				 */
 				.name		= "RGB-RAW",
 				.data_sets	= 4,
--- a/sensors/ev3_uart_sensor_defs.c
+++ b/sensors/ev3_uart_sensor_defs.c
@@ -156,18 +156,37 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
 			},
 			[5] = {
 				/**
-				 * .. [#lego-ev3-color-mode5] This mode is not
-				 *    usable. When in ``COL-CAL`` mode, the color
+				 * .. [#lego-ev3-color-mode5] This is a factory
+				 *    calibration mode. Writing
+				 *    ``LEGO-FAC-CAL-1`` directly to the sensor
+				 *    shortly after switching to this mode will
+				 *    trigger hardware recalibration. Then,
+				 *    after circa 500 ms, the new calibration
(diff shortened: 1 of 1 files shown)


sensors/ev3_uart_sensor_defs.c
```diff
@@ -86,25 +86,67 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
+				 *    on and off and measuring the current
+				 *    flowing through a bypass capacitor. It is
+				 *    not possible to measure true ambient light
+				 *    this way, but the LED off measurement
+				 *    works as a reference for the LED on
+				 *    measurement. Higher photodiode
+				 *    illumination will cause lower ADC
+				 *    measurements., so to obtain reflected
+				 *    light intensity similar to
+				 *    ``COL-REFLECT``, it is necessary to
```
dlech: This is markdown, so single backtick should be sufficient.

sensors/ev3_uart_sensor_defs.c
```diff
@@ -86,25 +86,67 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
+				 * @description: Raw Reflected Light - sets LED color to red
+				 * @value0: ADC reading for LED on (reflection) (0 to 1023)
+				 * @value0_footnote: [#lego-ev3-color-mode3-values]_
+				 * @value1: ADC reading for LED off (reference) (0 to 1023)
+				 * @value1_footnote: [#lego-ev3-color-mode3-values]_
 				 */
 				.name		= "REF-RAW",
 				.data_sets	= 2,
 				.data_type	= LEGO_SENSOR_DATA_S16,
+				.units		= "ADC",
```
dlech: This should exactly match what the sensor sends over the UART during initialization. My notes show this mode does not have units. Have you seen different?

sensors/ev3_uart_sensor_defs.c
```diff
@@ -86,25 +86,67 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
 				/**
+				 * .. [#lego-ev3-color-mode4-value012] These
+				 *    values represent the difference between
+				 *    the raw sensor ADC measurements when the
+				 *    LEDs are off and when they are on. They
+				 *    work similar to the ``COL-REFLECT`` values
+				 *    but they are not precalibrated.
+				 *
+				 * .. [#lego-ev3-color-mode4-value3] This value
+				 *    is similar to the REF-RAW ambient
```
dlech: Backticks around `REF-RAW` to be consistent?

sensors/ev3_uart_sensor_defs.c
```diff
@@ -86,25 +86,67 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
-				 * @value1: Green (0 to 1020???)
-				 * @value2: Blue (0 to 1020???)
-				 * @value3: Ambient? (0 to 1020???)
+				 * @value0: Reflected red light intensity (0 to 1023)
+				 * @value0_footnote: [#lego-ev3-color-mode4-value012]_
+				 * @value1: Reflected green light intensity (0 to 1023)
+				 * @value1_footnote: [#lego-ev3-color-mode4-value012]_
+				 * @value2: Reflected blue light intensity (0 to 1023)
+				 * @value2_footnote: [#lego-ev3-color-mode4-value012]_
+				 * @value3: ADC reading for LEDs off (reference) (0 to 1023)
```
dlech, line 150: I've been wondering if I should revert https://github.com/ev3dev/lego-linux-drivers/commit/3f7760ec7be41813a55dfd911a529606b38e1538. The sensor itself says that this mode only has 3 values. I got the idea about the 4th value from Dexter Industries BrickPi python library. But I've been wondering it this is actually a useful value or just leftover data in a buffer. UART sensors can only send data in sizes that are a power of 2, so that could explain why the extra data is sent even if it was never meant to be used.

Have you actually tried the ev3dev-buster kernel to see if this value seems similar to the same value in the RAW-REF mode?

sensors/ev3_uart_sensor_defs.c
```diff
@@ -86,25 +86,67 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
+				 * @value3_footnote: [#lego-ev3-color-mode4-value3]_
 				 */
 				.name		= "RGB-RAW",
 				.data_sets	= 4,
 				.data_type	= LEGO_SENSOR_DATA_S16,
+				.units		= "ADC",
```
dlech: Same here, this mode does not have units.

sensors/ev3_uart_sensor_defs.c
```diff
@@ -113,13 +155,19 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
+				 *    this will likely erase the
+				 *    factory-provided calibration.
 				 *
 				 * @name_footnote: [#lego-ev3-color-mode5]_
 				 * @description: Calibration ??? - sets LED color to red, flashing every 4 seconds, then goes continuous
-				 * @value0: ???
-				 * @value1: ???
-				 * @value2: ???
-				 * @value3: ???
+				 * @value0: ??? (0 to 1023???)
```
dlech: According to my notes, this mode has 65535 reported as the max value (both raw and SI units).

dlech (commented):
Thanks for the PR. I made a few comments/questions to have a look at.

sensors/ev3_uart_sensor_defs.c
JakubVanek: No, I haven't seen this, I thought this is only for documentation purposes. I will remove it.

sensors/ev3_uart_sensor_defs.c
JakubVanek: Yes, that is consistent with LEGO's `typedata.rcf`, I was guessing that this is coming from ADC, so >1023 wouldn't make sense. I will remove the change.

sensors/ev3_uart_sensor_defs.c
JakubVanek: Yup, I will remove it.

sensors/ev3_uart_sensor_defs.c
JakubVanek: I haven't tried this on ev3dev, but on stock firmware with c4ev3 it seemed to work similarly. I will recheck it. There is a simple test for this - if this value briefly changes when the intensity of external lighting illuminating the sensor changes, then this should be the ADC measurement when no LEDs are on.

I think that this value could still be useful - while one cannot measure steady ambient light, one could measure changes in ambient light this way.

sensors/ev3_uart_sensor_defs.c
JakubVanek: Yes, I will add them.

sensors/ev3_uart_sensor_defs.c
JakubVanek: I have found double backticks in the surrounding documentation (for example around line 313):
```
 * .. [#lego-ev3-gyro-mode0-value0-overflow] If
 *    you spin around too many times in ``GYRO-ANG``
 *    or ``GYRO-G&A`` mode, it will get stuck at
 *    32767 or overflow through -32768 depending
 *    on when the sensor was manufactured.
```

sensors/ev3_uart_sensor_defs.c
JakubVanek: On the stock firmware the values indeed exceed 1023, so the 65535 upper bound is likely to be true.

sensors/ev3_uart_sensor_defs.c
```diff
@@ -86,36 +86,87 @@ const struct ev3_uart_sensor_info ev3_uart_sensor_defs[] = {
...
+				 * @description: Raw Reflected Light - sets LED color to red
+				 * @value0: ADC reading for LED on (reflection) (0 to 1023)
+				 * @value0_footnote: [#lego-ev3-color-mode3-values]_
+				 * @value1: ADC reading for LED off (reference) (0 to 1023)
+				 * @value1_footnote: [#lego-ev3-color-mode3-values]_
 				 */
 				.name		= "REF-RAW",
 				.data_sets	= 2,
 				.data_type	= LEGO_SENSOR_DATA_S16,
+				.units		= "",
```
dlech: I don't think we need to add the units field here. This file already has the convention of using the default value for omitted fields.

sensors/ev3_uart_sensor_defs.c
dlech: OK, if it seems useful, then let's keep it. :smile:

JakubVanek: I have tried it again on C4EV3 (I don't have a SD card by me... :smile:) and the fourth value indeed behaves like the ambient of REF-RAW. I think this value can be useful and it's nice that ev3dev would provide it out of the box.

Without changes in external lighting, the values move around 670. When I move my phone's flashlight in front of the sensor, it temporarily moves to zero and the returns back. When I remove the light, the value jumps up and then returns again.

sensors/ev3_uart_sensor_defs.c
JakubVanek: Fixed

sensors/ev3_uart_sensor_defs.c
JakubVanek: Added 65535 as the upper bound :+1:

dlech:
merged, thanks!

PR #79, opened 2020-08-09
fix: change UART message buffer size to 35 bytes

JakubVanek:
Previously, the buffer was only 34 bytes large. This would overflow the buffer if an INFO message with 32 byte payload was received. The problem is the INFO command byte that is not counted to the payload size.

I haven't tested the change yet, but I will try to do so.

### DIFF
--- a/sensors/ev3_uart_sensor_ld.c
+++ b/sensors/ev3_uart_sensor_ld.c
@@ -66,7 +66,8 @@
 #endif
 
 #define EV3_UART_MAX_DATA_SIZE		32
-#define EV3_UART_MAX_MESSAGE_SIZE	(EV3_UART_MAX_DATA_SIZE + 2)
+/* extra bytes for: main command byte, INFO command byte, final checksum */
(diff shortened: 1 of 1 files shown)


JakubVanek:
The change compiles and the color sensor works with it on x86_64 laptop, hopefully nothing broke.

sensors/ev3_uart_sensor_ld.c
```diff
@@ -66,7 +66,8 @@
 #endif
 
 #define EV3_UART_MAX_DATA_SIZE		32
-#define EV3_UART_MAX_MESSAGE_SIZE	(EV3_UART_MAX_DATA_SIZE + 2)
+// extra bytes for: main command byte, INFO command byte, final checksum
```
dlech: Could you please change this to `/* */` style comments (it is the Linux kernel style.

dlech (commented):
Nice catch!

dlech (changes requested):
x

JakubVanek:
Thanks! I have now changed the style of the comment.

PR #80, opened 2020-08-09
feat: add RGB-RAW checksum calculation

JakubVanek:
Firmware inside the sensor XORs the previous byte at position 10 into the new checksum (going to position 10 too). By keeping track of the last byte value, we can predict the checksum value.

I have not yet tested the change, but I will try to do so.

### DIFF
--- a/sensors/ev3_uart_sensor_ld.c
+++ b/sensors/ev3_uart_sensor_ld.c
@@ -177,6 +177,8 @@ enum ev3_uart_info_flags {
  * 	mode in case of a reconnect.
  * @type_id: Type id returned by the sensor
  * @new_mode: The mode requested by set_mode.
+ * @last_byte_10: Last received value of 10th byte in any large enough message.
+ * 	This is used for calculating checksum byte for RGB-RAW DATA messages.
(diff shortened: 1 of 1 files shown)


JakubVanek:
The sensor now times out, something is wrong with this.

JakubVanek:
I think that my idea of what is happening inside the sensor was wrong. The code in question is here: https://github.com/JakubVanek/lms2012-stuff/blob/c50c019de2fa1074b0d5fda6558f412eeb99fff4/ev3color/program.c#L488-L518

The problem is that we cannot observe all transmit buffers, just the ones that were created when the UART tx queue was empty. Given that the sensor runs at 57600 baud and it needs to transmit 10 bytes, the sending time is 1.38 ms. This is longer than the 1 ms loop time maintained at the beginning of `update()`. So if the RGB values change in two consecutive milliseconds, the second change won't send the data message to the brick and the brick then won't have the correct byte 10 for checksum correction of future messages.

dlech:
Too bad a better workaround does not seem possible. Still very useful to have a better understanding of how the color sensor works though.

PR #81, opened 2020-08-09
Add data "age" attribute to sensor sysfs

JakubVanek:
I would like to propose a workaround for https://github.com/ev3dev/ev3dev/issues/1401 . A new sysfs attribute is exported: `bin_data_age`. It contains the number of milliseconds elapsed since the raw data received from given sensor last changed. This enables userspace to detect conditions like ultrasonic sensor not being able to measure distance at the current position. Thanks to `mode_info` separation present in current drivers, this also works across mode switches - if the same stale data are received after switch, the timestamp of last data update is not reset.

I think the naming (`bin_data_age`) is a bit unfortunate, but I don't know a better alternative. "Age" implies that the sensor did not send any new data during this time, but this may not be true. Sensor may still be sending data, but if they are the same as the old values, "age" is not reset to zero. I would like `bin_data_msec_since_last_change` more, but that is likely too long.

This patch also can't report age of individual values. One use I currently see is for detecting freezes on individual EV3 IR seeker channels. This would mean exporting age for each value separately, which may (or may not) be what is best for ev3dev.

The EV3 color sensor needs a different workaround - here the sensor sends new data, but they aren't accurate yet.

### DIFF
--- a/brickpi/brickpi_i2c_sensor.c
+++ b/brickpi/brickpi_i2c_sensor.c
@@ -76,6 +76,7 @@ static int brickpi_i2c_sensor_set_mode(void *context, u8 mode)
 		return err;
 
 	lego_port_set_raw_data_ptr_and_func(port, mode_info->raw_data, size,
+					    &mode_info->last_changed_time,
 					    NULL, NULL);
 
 	return 0;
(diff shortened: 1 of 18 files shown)


JakubVanek:
I have tested this only on x86_64 Linux laptop, but as the changes affect port code, I have to test it on the EV3 too. However, I would like to ask for help with testing on other platforms (brickpi, pistorms, wedo) to which I don't have access currently.

JakubVanek:
I have to test it tomorrow against `ev3dev-stretch`, I cannot get UART sensors to work even on stock `ev3dev-buster`.

JakubVanek:
The https://github.com/JakubVanek/ev3devdrv-private-dev/tree/feature/dataage-stretch branch (just ev3dev-stretch with this PR cherry-picked) appears to work well under EV3 with gyro, color, sonic and touch sensors.

dlech:
This is an interesting idea. I will have to think about it for a while (both the specific change and the bigger future of ev3dev in general).

PR #82, opened 2020-08-10
sensors/ev3_uart_sensor_ld: fix RGB-RAW condition check

JakubVanek:
Previously, checksum was ignored for any mode 4. This fixes it so that only color sensor mode 4 is removed from checksum calculation.


dlech:
Nice catch, thanks!

PR #83, opened 2020-08-10
sensors/ev3_uart_sensor_ld: sleep in ambient switch

JakubVanek:
EV3 Color Sensor produces invalid first few samples when a switch between COL-AMBIENT and other mode is performed. This PR adds a 10 ms `msleep()` after a mode switch if the switch is from/to COL-AMBIENT.

This works around the other part of https://github.com/ev3dev/ev3dev/issues/1401.


JakubVanek:
It seems that this successfully hides the issue on ev3dev-stretch when using pybricks-repl from repositories.

JakubVanek:
Hmm, it seems that https://github.com/pybricks/support/issues/14 was already fixed in the latest stretch pybricks version, so the effect was not due to this change.

JakubVanek:
Hmmm, my theory of how this occurs is probably wrong, https://github.com/pybricks/support/issues/14#issue-599780116 should work normally and it doesn't.

PR #84, opened 2022-01-23
changing HiTechnic IrSeeker V2 docs based on #1525

Eisverygoodletter:
changing some comments based off https://github.com/ev3dev/ev3dev/issues/1525

### DIFF
--- a/sensors/nxt_i2c_sensor_defs.c
+++ b/sensors/nxt_i2c_sensor_defs.c
@@ -772,7 +772,7 @@ const struct nxt_i2c_sensor_info nxt_i2c_sensor_defs[] = {
 				 *    =======  =============
 				 *
 				 * @description: Direction (unmodulated)
-				 * @value0: Direction (0 to 9)
+				 * @value0: Direction (0 to 9) These directions do not directly corresponds to the DC-ALL mode's 5 sensor values. the even number directions are calculated in the sensor (may have interference from sunlight)
 				 * @value0_footnote: [#ht-nxt-ir-seek-v2-mode0-value0]_
 				 */
 				.name = "DC",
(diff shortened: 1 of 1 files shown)


dlech:
Did you actually see values of 130? Or is it possible that the max is 127? The latter would make sense if they are using a signed 8-bit value.

Eisverygoodletter:
I actually measured a value of 170 when I put the IR ball really close to the sensor and orient the IR ball so that the light sources point directly at the sensor. https://www.robotshop.com/media/files/pdf/introduction-nsk1042.pdf says that it should be a byte (8 bits), so I think they may be using an unsigned 8-bit value.

dlech:
> so I think they may be using an unsigned 8-bit value.

I think so too, so lets make the range 0-255.

Eisverygoodletter:
Ok, but should I leave a note like "80 signal strength is approximately N cm distance "?

dlech:
I'm not sure how much that depends on the environment, especially with DC mode.

If you are quite sure that is accurate and most people will get the same results, we can add a footnote for it.

Eisverygoodletter:
Maybe let's add a footnote saying that values may not reach as high as 255 and is dependent on IR source strength and many other factors

dlech:
Sounds good.

Eisverygoodletter:
Here's a temporary build of the docs that show how the docs look like after the changes. https://irseeker-temporary-lego-linux-drivers.readthedocs.io/en/latest/sensor_data.html#hitechnic-nxt-irseeker-v2 Do you think that I should split the text into more footnotes, or should I keep it clean in 1 or 2 footnotes?

dlech:
I think it is fine the way it is.

Eisverygoodletter:
Then I think that that's pretty much all the changes done, ready to merge

dlech:
Thanks!

